PR #26, opened 2021-06-01
create hw_event_plugin

jzding:

### DIFF
--- a/plugins/hw_event/hw_event_plugin.go
+++ b/plugins/hw_event/hw_event_plugin.go
@@ -1,0 +1,207 @@
+// Copyright 2021 The Cloud Native Events Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"bytes"
+	"crypto/tls"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+	"sync"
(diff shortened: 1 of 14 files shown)


plugins/hw_event/hw_event_plugin.go
```diff
@@ -0,0 +1,207 @@
...
+)
+
+var (
+	resourceAddress string        = "/hw-event/cpu"
+	eventInterval   time.Duration = time.Second * 5
+	apiPath         string        = "/api/cloudNotifications/v1/"
+	apiAddr         string        = "localhost:8080"
+	localAPIAddr    string        = "localhost:9088"
+	localAPIPath    string        = "ack/event"
+	hwSubUrl        string        = "https://10.19.109.249/redfish/v1/EventService/Subscriptions/"
```
aneeshkp, line 46: IP address 10.19.109.249  is hard coded here, make it configurable

plugins/hw_event/hw_event_plugin.go
```diff
@@ -0,0 +1,207 @@
...
+		log.Errorf("failed to create a publisher %v", err)
+		return err
+	}
+	log.Printf("Created publisher %v", pub)
+
+	// once the publisher response is received, create a transport sender object to send events.
+	v1amqp.CreateSender(scConfig.EventInCh, pub.GetResource())
+
+	// serve ack/event event response
+	go server()
```
aneeshkp, line 89: Do we need a webserver here in the plugin? you can use default rest API  event ack route

plugins/hw_event/hw_event_plugin.go
```diff
@@ -0,0 +1,207 @@
...
+
+// createHwEventSubscription creates a subscription for Redfish HW events
+func createHwEventSubscription(eventTypes []string) (int, []byte) {
+
+	header := HttpHeader{ContentType: "application/json", ContentLanguage: "en-US"}
+	data := HwSubPayload{
+		Protocol: "Redfish",
+		Context:  "any string is valid",
+		// must be https
+		Destination: "https://localhost:8080/api/cloudNotifications/v1/create/event",
```
aneeshkp, line 104: This cannot be localhost, you need to add a route and use that route address. Because redfish will call externally. I think it makes perfect sense now to have a separate HW plugin web server and have it exposed externally

PR #31, opened 2021-06-21
Ptp metrics

aneeshkp:
ptp plugin will read data from the socket and generate events and metrics.


plugins/ptp_operator/metrics/metrics.go
```diff
@@ -0,0 +1,381 @@
...
+	replacer := strings.NewReplacer("[", " ", "]", " ", ":", " ")
+	output := replacer.Replace(msg)
+	fields := strings.Fields(output)
+	if len(fields) < 2 {
+		glog.Errorf("ignoring log:log  is not in required format processname[time]: [iface]")
+		return
+	}
+	processName := fields[0]
+	iface := fields[2]
+	if strings.Contains(output, " max ") { // this get generated in case -u is passed an option to phy2sys opts
```
atyronesmith, line 140: Are the " max " spaces needed or can you just look for max?

plugins/ptp_operator/metrics/metrics.go
```diff
@@ -0,0 +1,381 @@
...
+		offsetFromMaster, maxOffsetFromMaster, frequencyAdjustment, delayFromMaster := extractSummaryMetrics(processName, output)
+		UpdatePTPMetrics(processName, iface, offsetFromMaster, maxOffsetFromMaster, frequencyAdjustment, delayFromMaster)
+	} else if strings.Contains(output, " offset ") { // dispatcher calls metrics for phy2sys
```
atyronesmith, line 143: same here, need spaces?

plugins/ptp_operator/metrics/metrics.go
```diff
@@ -0,0 +1,381 @@
...
+		offsetFromMaster, maxOffsetFromMaster, frequencyAdjustment, delayFromMaster, clockState := extractRegularMetrics(processName, output)
+		// update stats
+		if clockState == "" {
+			return
+		}
+		var s *Stats
+		var found bool
+		if s, found = m.Stats[iface]; !found {
```
atyronesmith, line 151: Is a lock needed?  Is there another other thread that is modifying the the Stats map?

plugins/ptp_operator/metrics/metrics.go
```diff
@@ -0,0 +1,381 @@
...
+		glog.Warningf("no delay from master process %s out of sync", processName)
+	}
+
+	return
+}
+
+func extractRegularMetrics(processName, output string) (offsetFromMaster, maxOffsetFromMaster, frequencyAdjustment, delayFromMaster float64, clockState ceevent.SyncState) {
+	// remove everything before the rms string
+	// This makes the out to equals
+	output = strings.Replace(output, "path", "", 1)
```
atyronesmith, line 268: why do you need to remove "path"?

plugins/ptp_operator/metrics/metrics.go
```diff
@@ -0,0 +1,381 @@
...
+	}
+
+	if len(fields) >= 7 {
+		delayFromMaster, err = strconv.ParseFloat(fields[6], 64)
+		if err != nil {
+			glog.Errorf("%s failed to parse delay from master output %s error %v", processName, fields[6], err)
+		}
+	} else {
+		// If there is no delay from master this mean we are out of sync
+		glog.Warningf("no delay from master process %s out of sync", processName)
```
atyronesmith, line 313: This might spam the logs while you are out of lock.  Could you add a state variable such that you only print on going out of sync?  Also, should this information be incorporated into the state calculation?

plugins/ptp_operator/metrics/metrics.go
```diff
@@ -0,0 +1,381 @@
...
+	return
+}
+
+// ExtractPTP4lEvent ... extract event form ptp4l logs
+func extractPTP4lEventState(processName, output string) (clockState ceevent.SyncState) {
+	// This makes the out to equals
+	//phc2sys[187248.740]:[ens5f0] CLOCK_REALTIME phc offset        12 s2 freq   +6879 delay    49
+	// phc2sys[187248.740]:[ens5f1] CLOCK_REALTIME phc offset        12 s2 freq   +6879 delay    49
+	//ptp4l[3535499.740]: [ens5f0] master offset         -6 s2 freq   -1879 path delay        88
+	if strings.Contains(output, "INITIALIZING to LISTENING on INIT_COMPLETE") {
```
atyronesmith, line 325: seems like you could look for shorter stings or event words here...

plugins/ptp_operator/metrics/metrics.go
```diff
@@ -0,0 +1,381 @@
...
+		clockState = ceevent.FREERUN
+	} else if strings.Contains(output, "LISTENING to UNCALIBRATED on RS_SLAVE") {
+		clockState = ceevent.FREERUN
+	} else if strings.Contains(output, "UNCALIBRATED to SLAVE on MASTER_CLOCK_SELECTED") {
+		clockState = ceevent.FREERUN
+	} else if strings.Contains(output, "SLAVE to FAULTY on FAULT_DETECTED") {
```
atyronesmith, line 331: could just look for FAULT_DETECTED?

plugins/ptp_operator/metrics/metrics.go
```diff
@@ -0,0 +1,381 @@
...
+
+// ExtractPTP4lEvent ... extract event form ptp4l logs
+func extractPTP4lEventState(processName, output string) (clockState ceevent.SyncState) {
+	// This makes the out to equals
+	//phc2sys[187248.740]:[ens5f0] CLOCK_REALTIME phc offset        12 s2 freq   +6879 delay    49
+	// phc2sys[187248.740]:[ens5f1] CLOCK_REALTIME phc offset        12 s2 freq   +6879 delay    49
+	//ptp4l[3535499.740]: [ens5f0] master offset         -6 s2 freq   -1879 path delay        88
+	if strings.Contains(output, "INITIALIZING to LISTENING on INIT_COMPLETE") {
+		clockState = ceevent.FREERUN
+	} else if strings.Contains(output, "LISTENING to UNCALIBRATED on RS_SLAVE") {
```
atyronesmith, line 327: Contains(output,"RS_SLAVE") ... ?

atyronesmith (commented):
looks efficient!  Some suggests included...

plugins/ptp_operator/metrics/metrics.go
aneeshkp: because ptp4l has only `delay`  and phc2sys has `path delay`, so it makes the output equal

plugins/ptp_operator/metrics/metrics.go
aneeshkp: Yes , Will change to that

plugins/ptp_operator/ptp_operator_plugin_test.go
```diff
@@ -164,7 +163,7 @@ func ProcessInChannel() {
 		select {
 		case d := <-scConfig.EventInCh:
 			if d.Type == channel.LISTENER {
-				log.Printf("amqp disabled,no action taken: request to create listener  address %s was called,but transport is not enabled", d.Address)
+				log.Printf("1amqp disabled,no action taken: request to create listener address %s was called,but transport is not enabled", d.Address)
```
jzding, line 166: typo? `1amqp`

plugins/ptp_operator/metrics/metrics.go
```diff
@@ -0,0 +1,381 @@
...
+
+func (s *Stats) addValue(val float64) {
+
+	oldMean := s.mean
+
+	if s.num == 0 || s.max < val {
+		s.max = val
+	}
+	if s.num == 0 && s.min > val {
+		s.mean = val
```
jzding, line 90: `s.min` ?

plugins/ptp_operator/metrics/metrics.go
aneeshkp: Fixed

plugins/ptp_operator/metrics/metrics.go
aneeshkp: fixed

plugins/ptp_operator/metrics/metrics.go
aneeshkp: Fixed added new state

plugins/ptp_operator/metrics/metrics.go
aneeshkp: Fixed: Used RegEx. Please review

plugins/ptp_operator/metrics/metrics.go
aneeshkp: Using RegEx, please review

plugins/ptp_operator/metrics/metrics.go
aneeshkp: using RegEx , please review

cmd/main.go
```diff
@@ -86,10 +86,13 @@ func main() {
 		BaseURL:    nil,
 	}
 
+
+	metricServer(metricsAddr)
```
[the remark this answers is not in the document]
aneeshkp: It was a merge issue. I cleaned it up.

cmd/main.go
```diff
@@ -85,9 +85,18 @@ func main() {
 		BaseURL:    nil,
 	}
 
-	wg := sync.WaitGroup{}
-	wg.Add(1)
 	metricServer(metricsAddr)
+	wg := sync.WaitGroup{}
+	c := make(chan os.Signal)
```
jzding, line 90: add a buffer to fix lint error.
`c := make(chan os.Signal, 1)`

plugins/ptp_operator/metrics/metrics.go
```diff
@@ -0,0 +1,381 @@
...
+		glog.Errorf("%s failed to parse output %s: unexpected number of fields", processName, output)
+		return
+	}
+
+	offsetFromMaster, err := strconv.ParseFloat(fields[1], 64)
+	if err != nil {
+		glog.Errorf("%s failed to parse offset from master output %s error %v", processName, fields[1], err)
+	}
+
+	maxOffsetFromMaster, err = strconv.ParseFloat(fields[1], 64)
```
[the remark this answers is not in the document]
aneeshkp: We are parsing two logs ptp4l and phc2sys . The ptpp4l doesn't have a max offset. Only phc2sys logs will have max offset. So this is by default used by ptp Linux daemon. 
if -u is passed.
But for events you should not pass -u to phc2sys, in that case, we lose maxOffset. Currently I am calculating max offset   to retain metrics without the -u option

plugins/ptp_operator/ptp_operator_plugin_test.go
aneeshkp: Thanks.fixed

plugins/ptp_operator/metrics/metrics.go
aneeshkp: No, Also I need to remove the grouting call in the socket read function.
```
for {
		fd, err := l.Accept()
		if err != nil {
			log.Errorf("accept error: %s", err)
		} else {
			go processMessages(fd)
		}
	}
```

plugins/ptp_operator/metrics/metrics.go
aneeshkp: oops. fixed
```
if s.num == 0 && s.min > val {
		s.min = val
	}
	```

File: cmd/main.go (hunk 12 above)
Comment (aneeshkp) [reply cmd/main.go:90]: fixed

File: plugins/ptp_operator/metrics/metrics.go [hunk 14]
```diff
@@ -0,0 +1,381 @@
...
+	indx := strings.Index(output, "rms")
+	output = output[indx:]
+	fields := strings.Fields(output)
+
+	if len(fields) < 5 {
+		glog.Errorf("%s failed to parse output %s: unexpected number of fields", processName, output)
+		return
+	}
+
+	offsetFromMaster, err := strconv.ParseFloat(fields[1], 64)
```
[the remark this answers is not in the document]
aneeshkp: good idea, but that will not make it readable, since we are splitting the strings. I will leave it at index so it is easy to see the positioning from the string

pkg/common/common.go
```diff
@@ -148,7 +148,7 @@ func PublishEvent(scConfig *SCConfiguration, e ceevent.Event) error {
 		log.Errorf("error posting ptp events %v", err)
 		return err
 	}
-	log.Errorf("published ptp event %s", e.String())
+	log.Infof("published ptp event %s", e.String())
```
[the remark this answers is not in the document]
aneeshkp: Fixed

plugins/ptp_operator/metrics/metrics.go
jzding: shouldn't it be `||` here?
```
if s.num == 0 || s.min > val {
		s.min = val
}
```

plugins/ptp_operator/metrics/metrics.go
```diff
@@ -0,0 +1,421 @@
...
+// ExtractPTP4lEvent ... extract event form ptp4l logs
+func extractPTP4lEventState(processName, output string) (clockState ceevent.SyncState) {
+	// This makes the out to equals
+	//phc2sys[187248.740]:[ens5f0] CLOCK_REALTIME phc offset        12 s2 freq   +6879 delay    49
+	// phc2sys[187248.740]:[ens5f1] CLOCK_REALTIME phc offset        12 s2 freq   +6879 delay    49
+	//ptp4l[3535499.740]: [ens5f0] master offset         -6 s2 freq   -1879 path delay        88
+
+	/*
+		"INITIALIZING to LISTENING on INIT_COMPLETE"
+		"INITIALIZING to LISTENING on INIT_COMPLETE"
```
jzding, line 406: duplicated

plugins/ptp_operator/metrics/metrics.go
```diff
@@ -0,0 +1,421 @@
...
+		"LISTENING to UNCALIBRATED on RS_SLAVE"
+		"UNCALIBRATED to SLAVE on MASTER_CLOCK_SELECTED"
+		"SLAVE to FAULTY on FAULT_DETECTED"
+		"LISTENING to FAULTY on FAULT_DETECTED"
+		"FAULTY to LISTENING on INIT_COMPLETE"
+		"FAULTY to SLAVE on INIT_COMPLETE"
+		"SLAVE to UNCALIBRATED on SYNCHRONIZATION_FAULT"
+	*/
+	var rePTPStatus = regexp.MustCompile(`INITIALIZING|LISTENING|UNCALIBRATED|FAULTY|FAULT_DETECTED|SYNCHRONIZATION_FAULT`)
```
jzding, line 415: patterns like `FAULTY` looks too general to me.

plugins/ptp_operator/metrics/metrics.go
aneeshkp: removed

plugins/ptp_operator/metrics/metrics.go
aneeshkp: That's a fixed set of values coming from the PTP logs.  All values will transition to FREERUN. After testing in real PTP  labs if we see the difference then we can enhance it. Also, I have to add a configurable threshold variable.

plugins/ptp_operator/metrics/metrics.go
atyronesmith: ok.

plugins/ptp_operator/metrics/metrics.go
```diff
@@ -0,0 +1,420 @@
...
+
+func (s *Stats) addValue(val float64) {
+
+	oldMean := s.mean
+
+	if s.max < val {
+		s.max = val
+	}
+	if s.min > val {
+		s.min = val
```
jzding, line 113: If s.min is not initialized then this will always make s.min = 0. I think it makes sense to keep `s.num == 0 || ` here.

plugins/ptp_operator/metrics/metrics.go
```diff
@@ -0,0 +1,420 @@
...
+func (s *Stats) getMaxAbs() float64 {
+	if s.max > s.mean {
+		return s.max
+	}
+	return s.min
+
+}
+
+// reset status
+func (s *Stats) reset() { //nolint:unused
```
jzding, line 138: Does `s.min` need to be reset?

plugins/ptp_operator/metrics/metrics.go
aneeshkp: in Go  variables declared without an initial value are set to their zero values.. so here it will be 0.0

jzding: Without `s.num == 0 || `, if all values in s are greater than 0, the min is still 0. Is this intended?

aneeshkp: right ,I  will add it back. we need to capture the first value when its locked as min

plugins/ptp_operator/metrics/metrics.go
aneeshkp: Added and also  making reset call when its not in Locked state as per PTP4l log

jzding (dismissed):

jzding (approved):

jzding (dismissed):

jzding (approved):

plugins/ptp_operator/metrics/metrics.go
```diff
@@ -239,28 +265,58 @@ func (p *PTPEventManager) ExtractMetrics(msg string) {
...
+	switch clockState {
+	case ceevent.LOCKED:
+		switch lastClockState {
+		case ceevent.FREERUN: //last state was already sent for FreeRUN n, but if its within then send again with new state
+			if offsetFromMaster != 0 && (offsetFromMaster < p.maxOffsetThreshold && offsetFromMaster > p.minOffsetThreshold) { // within range
+				p.publishEvent(clockState, iface) // change to locked
+				iStats.lastClockState = clockState
+				iStats.lastClockStateCount = 1
+			}
+		case ceevent.LOCKED: // last state was in sync , check if its pout of sync
```
jzding, line 280: typo `out of sync`

plugins/ptp_operator/metrics/metrics.go
```diff
@@ -239,28 +265,58 @@ func (p *PTPEventManager) ExtractMetrics(msg string) {
...
+				iStats.lastClockStateCount = 1
+			}
+		default: // not yet used states
+			p.publishEvent(clockState, iface) // change to locked
+			iStats.lastClockState = clockState
+			iStats.lastClockStateCount = 1
+			//do nothing already send event
+		}
+	case ceevent.FREERUN:
+		if offsetFromMaster != 0 && (offsetFromMaster < p.maxOffsetThreshold && offsetFromMaster > p.minOffsetThreshold) { // within range
```
jzding, line 307: would look cleaner to have utility functions withinRange(), outOfRange() and lockState()

PR #44, opened 2021-07-19
Added features to monitor ptpconfig configmap for changes in node prof…

aneeshkp:
PTP Thresholds like
 1.HoldoverTimeout
2. MaxOffsetThreshold
3. MinOffsetThreshold
are now set in ptp-operator CR called ptp-config. These values are converted into the ptpConfig config map.
The threshold values are then read by the sidecar by polling for changes every 30secs .
This adds that functionality to monitor ptpconfigmap for any changes and update the sidecar with threshold values.


plugins/ptp_operator/config/config.go
```diff
@@ -0,0 +1,216 @@
...
+	for _, profile := range l.NodeProfiles {
+		if _, found := l.EventThreshold[*profile.Interface]; !found {
+			l.SetDefaultPTPThreshold(*profile.Interface)
+		}
+		if t, found := l.EventThreshold[*profile.Interface]; found {
+			if profile.PtpClockThreshold.MaxOffsetThreshold != 0 {
+				t.MaxOffsetThreshold = profile.PtpClockThreshold.MaxOffsetThreshold
+			}
+			if profile.PtpClockThreshold.MinOffsetThreshold != 0 {
+				t.MinOffsetThreshold = profile.PtpClockThreshold.MinOffsetThreshold
```
jzding, line 108: Is it possible we want to update MinOffsetThreshold or HoldOverTimeout to `0`?

plugins/ptp_operator/config/config.go
```diff
@@ -0,0 +1,216 @@
...
+			if err != nil {
+				log.Errorf("error reading node profile: %v", nodeProfile)
+				continue
+			}
+			err = l.UpdateConfig(nodeProfilesJSON)
+			if err != nil {
+				log.Errorf("error updating the node configuration using the profiles loaded: %v", err)
+			}
+		case sig := <-closeCh:
+			log.Info("signal received, shutting down", sig)
```
jzding, line 203: Either change to `log.Infof` to add %arg for `sig`.

plugins/ptp_operator/config/config.go
```diff
@@ -0,0 +1,216 @@
...
+}
+
+// WatchConfigUpdate watch for ptp config update
+func (l *LinuxPTPConfUpdate) WatchConfigUpdate(nodeName string, closeCh chan struct{}) {
+	tickerPull := time.NewTicker(time.Duration(l.intervalUpdate) * time.Second)
+	defer tickerPull.Stop()
+	for {
+		select {
+		case <-tickerPull.C:
+			log.Infof("ticker pull for ptp profile updates")
```
jzding, line 181: Should be `log.Info` with no arguments. (or should we use `log.debug`?)

jzding (dismissed):

plugins/ptp_operator/config/config.go
aneeshkp: not with holdover time. it is secs and min is 1secs to the max can be anything.
the minOffsetThreshold can be zero, but zero is a perfect offset which is hard to achieve.
I could remove the check of min and max threshold

plugins/ptp_operator/config/config.go
aneeshkp: looks like our linter is broken, Usually, the linter should catch this type. I will fix this

plugins/ptp_operator/config/config.go
aneeshkp: looks like our linter is broken, Usually, the linter should catch this type. I will fix this

plugins/ptp_operator/config/config.go
aneeshkp: Changed the logic a bit to check for default values

plugins/ptp_operator/config/config.go
aneeshkp: Fixed

plugins/ptp_operator/config/config.go
aneeshkp: Fixed

jzding (approved):

PR #45, opened 2021-07-19
vendor dependencies

nishant-parekh:
Vendor dependencies to help ART productization.

Signed-off-by:Nishant Parekh nparekh@redhat.com


aneeshkp:
I think you need to set this to off in the main Dockerfile
https://github.com/redhat-cne/cloud-event-proxy/blob/main/Dockerfile#L2

hack/build-go.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
 set -eu
-
+make deps-update
```
aneeshkp, line 4: I think we should not call every time it builds. SO remove this

Makefile
```diff
@@ -38,6 +38,10 @@ else
 KUSTOMIZE=$(shell which kustomize)
 endif
```
aneeshkp, line 39: you need to test this. I guess it is still using GPATH 
you need to add go build-mod=vendor

aneeshkp (approved):
/lgtm

PR #55, opened 2021-07-23
Remove hw-event-plugin

jzding:


examples/manifests/consumer.yaml
```diff
@@ -49,8 +49,12 @@ spec:
           env:
             - name: HW_PLUGIN
```
aneeshkp, line 50: consumer should not have any Plugin valraibles

aneeshkp: only producer uses plugins

examples/manifests/hw-event-proxy.yaml
```diff
@@ -37,6 +37,10 @@ spec:
           env:
             - name: HW_PLUGIN
               value: "true"
+            - name: PTP_PLUGIN
```
aneeshkp, line 40: since this is hw specific manifests, you should remove other plugin variables from this yaml

examples/manifests/kustomization.yaml
```diff
@@ -16,7 +16,10 @@ replicas:
 images:
 - name: cloud-event-proxy
   newName: quay.io/aneeshkp/cloud-event-proxy
```
aneeshkp, line 18: I think we can soon change this to the quay.io/openshift/cloud-event-proxy

examples/manifests/kustomization.yaml
```diff
@@ -16,7 +16,10 @@ replicas:
...
+  newTag: latest
 - name: cloud-native-event-consumer
   newName: quay.io/aneeshkp/cloud-native-event-consumer
+  newTag: latest
 - name: cloud-native-event-producer
   newName: quay.io/aneeshkp/cloud-native-event-producer
```
aneeshkp, line 24: you don't need producer here anymore. The producer is cloud-event-proxy with hw-plugin enabled right ? You can delete the producer completely

pkg/common/common.go
```diff
@@ -163,7 +163,7 @@ func CreateHwEvent(pubSubID, eventType string, data hwevent.Data) (hwevent.Event
 	event := v1hwevent.CloudNativeEvent()
 	event.ID = pubSubID
 	event.Type = eventType
-	event.SetTime(types.Timestamp{Time: time.Now().UTC()}.Time)
+	event.SetTime(t)
```
aneeshkp, line 166: time should be time created, why passing the time? 
In publishEvent function you are creating the time not using time received 
`// Use time received to calculate lapse
	timeReceived := types.Timestamp{Time: time.Now().UTC()}.Time`

Also you can set the time after you received the  event object

plugins/hw_event/gen.go
```diff
@@ -0,0 +1,7 @@
+package main
```
aneeshkp, line 1: is this a required file?

plugins/hw_event/hw_event_plugin.go
```diff
@@ -189,9 +101,11 @@ func publishHwEvent(w http.ResponseWriter, r *http.Request) {
 	defer r.Body.Close()
 	bodyBytes, err := ioutil.ReadAll(r.Body)
 	if err != nil {
-		log.Errorf("error reading hw event %v", err)
+		log.Errorf("error reading hw event: %v", err)
 		return
 	}
+	// Use time received to calculate lapse
+	timeReceived := types.Timestamp{Time: time.Now().UTC()}.Time
```
aneeshkp, line 108: this is not time received, it should be time event created

plugins/hw_event/hw_event_plugin.go
```diff
@@ -200,10 +114,48 @@ func publishHwEvent(w http.ResponseWriter, r *http.Request) {
...
+		}
+	}
+
 	data := v1hwevent.CloudNativeData()
 	data.SetVersion("v1") //nolint:errcheck
 	data.SetData(&redfishEvent)
 
-	event, _ := common.CreateHwEvent(pub.ID, "HW_EVENT", data)
-	_ = common.PublishHwEvent(scConfig, event)
+	event, _ := common.CreateHwEvent(pub.ID, "HW_EVENT", data, timeReceived)
```
aneeshkp, line 134: instead of passing the time created, you can set the event object after it is received.
Also check for error  before setting

plugins/hw_event/hw_event_plugin.go
```diff
@@ -200,10 +114,48 @@ func publishHwEvent(w http.ResponseWriter, r *http.Request) {
...
+	_ = common.PublishHwEventViaAPI(scConfig, event)
+}
+
+func parseMessage(m hwevent.EventRecord) (hwevent.EventRecord, error) {
+	addr := "localhost:9999"
```
aneeshkp, line 139: can you explain why port is hardcoded? also move it to a constant

plugins/hw_event/gen.go
jzding: it is for generating code under pb, but it is not build into the binary.

plugins/hw_event/hw_event_plugin.go
jzding: This is temporary. I have a followup story to put message-parser in container, and will pass port number from environment variable.

pkg/common/common.go
jzding: updated to use timestamp from Redfish event

examples/manifests/consumer.yaml
jzding: updated

examples/manifests/hw-event-proxy.yaml
jzding: removed

pkg/common/common.go
aneeshkp: instead of passing the time why don't you set the time after you received the event object ?

plugins/hw_event/hw_event_plugin.go
```diff
@@ -15,60 +15,38 @@
...
 	v1pubsub "github.com/redhat-cne/sdk-go/v1/pubsub"
 	log "github.com/sirupsen/logrus"
 )
 
 var (
-	resourceAddress string   = "/hw-event"
-	hwSubURL        string   = "https://10.19.109.249/redfish/v1/EventService/Subscriptions/"
-	hwEventTypes    []string = []string{"StatusChange", "ResourceUpdated", "ResourceAdded", "ResourceRemoved", "Alert"}
-
+	resourceAddress string = "/hw-event"
```
aneeshkp, line 41: do you want to call this as hw-event?
may be the resource should have also node-name, so the subscribe can choose to subscribe to node it is interested of 
/cluster/node/%s/redfish/event
also have ability to subscribe to single HW events , for an example 
/cluster/node/%s/redfish/event/fan

pkg/common/common.go
aneeshkp: https://github.com/redhat-cne/cloud-event-proxy/blob/42dc03d5cd2416e27772bd9f60e803ac0a7db11f/plugins/hw_event/hw_event_plugin.go#L108
here create the time after you got the event object . That way it  will  be more accurate

plugins/hw_event/hw_event_plugin.go
aneeshkp: move the string to a constant

examples/manifests/kustomization.yaml
jzding: so no change for now?

examples/manifests/kustomization.yaml
jzding: removed

pkg/common/common.go
jzding: done. Set the event.Time when received Redfish event from webhook.

plugins/hw_event/hw_event_plugin.go
jzding: updated

plugins/hw_event/hw_event_plugin.go
jzding: updated

plugins/hw_event/hw_event_plugin.go
jzding: done

plugins/hw_event/hw_event_plugin.go
jzding: done at hw-event-proxy

examples/consumer/main.go
```diff
@@ -218,6 +221,9 @@ func getEvent(w http.ResponseWriter, req *http.Request) {
 	}
 	e := string(bodyBytes)
 	if e != "" {
+		if consumerType == HW {
```
aneeshkp, line 224: I believe you need this  to run consumer  for HW events

examples/manifests/hw-event-proxy.yaml
```diff
@@ -56,7 +56,7 @@ spec:
                 fieldRef:
                   fieldPath: status.podIP
             - name: AMQP_NAMESPACE
-              value: "amqp"
+              value: "amqp-interconnect"
```
aneeshkp, line 59: do you need hw-event-proxt.yaml

aneeshkp: This file can be deleted I think

examples/manifests/kustomization.yaml
aneeshkp: Nishanth has to fix that ,once he fix that we will do a PR to change all images to point to openshift quay

aneeshkp: but consumer (example) will still stay under private repo

examples/manifests/route.yaml
```diff
@@ -3,6 +3,7 @@ apiVersion: route.openshift.io/v1
 kind: Route
 metadata:
   name: hw-event-proxy
+  namespace: cloud-native-events
```
aneeshkp, line 6: do you need this route.yaml here ? it should be under hw-event-proxy

go.mod
```diff
@@ -5,9 +5,9 @@ go 1.15
 require (
 	github.com/cloudevents/sdk-go/v2 v2.4.1
 	github.com/prometheus/client_golang v1.11.0
-	github.com/redhat-cne/rest-api v0.0.0-20210628151454-c8f37f0a9c3e
-	github.com/redhat-cne/sdk-go v0.0.0-20210630032545-4072da5111ba
+	github.com/redhat-cne/rest-api v0.1.0
```
aneeshkp, line 8: Thanks for this fix. We  missed it in current main branch .

aneeshkp:
@jzding  Please change the PR name and add little context to this PR

examples/manifests/hw-event-proxy.yaml
aneeshkp: Delete this file

examples/consumer/main.go
jzding: yes just for monitoring the performance

examples/manifests/hw-event-proxy.yaml
jzding: deleted

examples/manifests/route.yaml
jzding: deleted

examples/manifests/kustomization.yaml
```diff
@@ -16,7 +14,7 @@ replicas:
 images:
```
aneeshkp, line 14: hw-event-proxy needs to be removed

examples/consumer/main.go
aneeshkp: also you need for QE to run manuall tests

examples/manifests/kustomization.yaml
jzding: done

PR #58, opened 2021-08-15
Bug 1993854:Multiple interface

aneeshkp:
This PR is to support Multi Interface for PTP boundary clock, which can have a single slave port and multiple master
ports,
Following metrics were added 
1.offset_from_system
2.max_offset_from_system
3.frequency_adjustment_from_system & frequency_adjustment_from_master
4.delay_from_master and delay_from system

Event reporting logic is changed . It will not report interface name of current slave, instead it will report CLOCK_REALTIME events for slave interface and new  Prometheus metrics is introduced to identify SLAVE/MASTER/PASSIVE and FAULTY status of the interfaces .
All events reported by PTP4L will generate an event. like interface role changes and faulty .
All sync state will be reported via phc2sys logs by comparing  offset against threshold
There will always be one slave and that will be reported as CLOCK_REALTIME in Prometheus metrics , and on every  new slave  role take over will update clock_realtime.
A new metrics were added to report interface roles .
A file watcher is implemented to watch for ptp4l config files and update the interface list .


plugins/ptp_operator/metrics/metrics.go
```diff
@@ -225,149 +308,288 @@ func (p *PTPEventManager) MockTest(t bool) {
...
+				}
 			} else {
-				p.Stats[iface].lastOffset = int64(offsetFromMaster)
-				p.Stats[iface].addValue(int64(offsetFromMaster)) // update off set when its in locked state and hold over only
+				for _, i := range ifaces {
+					p.Stats[i].lastOffset = int64(offsetFromMaster)
+					p.Stats[i].addValue(int64(offsetFromMaster)) // update off set when its in locked state and hold over only
+				}
 			}
 		case ceevent.HOLDOVER: // last state was holdover, since it is in locked now check if it qualifies to be locked
```
jzding, line 557: the comment followed can be deleted since it is contradict with the comment in the next line.

plugins/ptp_operator/metrics/metrics.go
```diff
@@ -225,149 +308,288 @@ func (p *PTPEventManager) MockTest(t bool) {
...
+				p.Stats[i].addValue(int64(offsetFromMaster))
+			}
 		}
 	default:
 		log.Warnf("%s unkown current ptp state %s", iface, clockState)
-		p.PublishEvent(clockState, offsetFromMaster, iface, channel.PTPEvent) // change to locked
-		p.Stats[iface].lastClockState = clockState
-		p.Stats[iface].lastOffset = int64(offsetFromMaster)
-		p.Stats[iface].reset() // reset when its free run
+		p.PublishEvent(clockState, offsetFromMaster, string(iface), channel.PTPEvent) // change to locked
```
jzding, line 579: comment says "change to locked" but the event published is unknown state (`clockState`).

plugins/ptp_operator/config/config.go
```diff
@@ -1,33 +1,58 @@
 package config
 
+// Copyright 2020 The Cloud Native Events Authors
```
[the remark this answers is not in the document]
aneeshkp: I will leave it 2020, it all over the place as 2020 . wont matter

plugins/ptp_operator/metrics/metrics.go
```diff
@@ -225,149 +308,288 @@ func (p *PTPEventManager) MockTest(t bool) {
...
+func (p *PTPEventManager) GetPTPConfig(configName types.ConfigName) *ptp4lconf.PTP4lConfig {
+
+	if _, ok := p.Ptp4lConfigInterfaces[configName]; ok && p.Ptp4lConfigInterfaces[configName] != nil {
+		return p.Ptp4lConfigInterfaces[configName]
+	}
+	pc := &ptp4lconf.PTP4lConfig{
+		Name: string(configName),
+	}
+	pc.Interfaces = []*ptp4lconf.PTPInterface{}
+	p.AddPTPConfig(configName, pc)
```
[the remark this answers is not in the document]
aneeshkp: This function is GetPTPConfig, if there is no config stored then it creates empty interfaces else this conditions satisfies and returns current stored config
`if _, ok := p.Ptp4lConfigInterfaces[configName]; ok && p.Ptp4lConfigInterfaces[configName] != nil {
		return p.Ptp4lConfigInterfaces[configName]
	}`
ptp-Linux daemon has a contract to send config name in the log section 

There is a function to validate configName  when first reading the logs
ptpConfigFileRegEx = regexp.MustCompile(`ptp4l.[0-9]*.config`)
//FindInLogForCfgFileIndex ...
`func FindInLogForCfgFileIndex(out string) int {
	match := ptpConfigFileRegEx.FindStringIndex(out)
	if len(match) == 2 {
		return match[0]
	}
	return -1
}
`

plugins/ptp_operator/metrics/metrics.go
```diff
@@ -225,149 +308,288 @@ func (p *PTPEventManager) MockTest(t bool) {
...
-		// update stats
-		if clockState == "" {
-			return
-		}
-		if _, found := p.Stats[iface]; !found {
-			p.Stats[iface] = NewStats()
+	configName := fields[2]
+	ptp4lCfg := p.GetPTPConfig(types.ConfigName(configName))
+
+	if ptp4lCfg == nil { //TODO: Use PMC to update port and roles
```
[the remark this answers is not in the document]
aneeshkp: good catch , wonder why linter didn't catch that ?
changed it to if len(ptp4lCfg.Interfaces) == 0

plugins/ptp_operator/metrics/metrics.go
```diff
@@ -225,149 +308,288 @@ func (p *PTPEventManager) MockTest(t bool) {
...
-		UpdatePTPMetrics(processName, iface, offsetFromMaster, maxOffsetFromMaster, frequencyAdjustment, delayFromMaster, "")
-	} else if strings.Contains(output, " offset ") {
-		offsetFromMaster, maxOffsetFromMaster, frequencyAdjustment, delayFromMaster, clockState := extractRegularMetrics(processName, output)
-		// update stats
-		if clockState == "" {
-			return
-		}
-		if _, found := p.Stats[iface]; !found {
-			p.Stats[iface] = NewStats()
+	configName := fields[2]
```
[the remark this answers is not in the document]
aneeshkp: done, added two constants

plugins/ptp_operator/config/config.go
```diff
@@ -1,33 +1,58 @@
...
-	// DefaultUpdateInterval for watching ptpconfig update
-	DefaultUpdateInterval = 45
+	// DefaultUpdateInterval for watching ptpconfigmap update
+	DefaultUpdateInterval = 60
 	// DefaultProfilePath  default ptp profile path
 	DefaultProfilePath       = "/etc/linuxptp"
 	maxOffsetThreshold int64 = 100 //in nano secs
 	minOffsetThreshold int64 = -100
 	holdoverTimeout    int64 = 5
+	ignorePtp4lSection       = "global"
```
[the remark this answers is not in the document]
aneeshkp: renamed to `ptp4lGlobalSection`

plugins/ptp_operator/metrics/metrics.go
aneeshkp: done

plugins/ptp_operator/metrics/metrics.go
aneeshkp: changed comment to unknown.
 this catches any state PTP produces that are unknown to us , since we are still exploring the ptp log events.

plugins/ptp_operator/config/config.go
```diff
@@ -75,66 +102,85 @@ func NewLinuxPTPConfUpdate() *LinuxPTPConfUpdate {
...
+			if profile.PtpClockThreshold.MaxOffsetThreshold > 0 { // has to be greater than 0 nano secs
+				threshold.MaxOffsetThreshold = profile.PtpClockThreshold.MaxOffsetThreshold
+			} else {
+				threshold.MaxOffsetThreshold = maxOffsetThreshold
+				log.Infof("maxOffsetThreshold %d has to be > 0, now set to default %d", profile.PtpClockThreshold.MaxOffsetThreshold, maxOffsetThreshold)
+			}
+			if profile.PtpClockThreshold.MinOffsetThreshold > threshold.MaxOffsetThreshold {
+				threshold.MinOffsetThreshold = threshold.MaxOffsetThreshold - 1 // make it one nano secs less than max
+				log.Infof("minOffsetThreshold %d has to be > maxOffsetThreshold, now set to one less than maxOfssetThreshold %d",
+					profile.PtpClockThreshold.MinOffsetThreshold, maxOffsetThreshold)
```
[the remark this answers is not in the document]
jzding: In `"now set to one less than maxOfssetThreshold %d"`, `%d` should take the current value `threshold.MaxOffsetThreshold`, not default value `maxOffsetThreshold` which is 100.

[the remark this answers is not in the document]
aneeshkp: you are right, thanks.Will fix the log message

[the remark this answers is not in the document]
aneeshkp: done

plugins/ptp_operator/ptp4lconf/ptp4lConfig.go
```diff
@@ -0,0 +1,233 @@
...
+			updatedConfig <- ptpConfig
+		}
+		for {
+			select {
+			case event, ok := <-w.fsWatcher.Events:
+				if !ok {
+					log.Error("could not read watcher events")
+					continue
+				}
+				log.Infof("event: %v", event)
```
jzding, line 157: are there a lot of events? if so might want to change to debugf.

plugins/ptp_operator/ptp4lconf/ptp4lConfig.go
```diff
@@ -0,0 +1,233 @@
...
+
+func readAllConfig(dir string) []*PtpConfigUpdate {
+	var ptpConfigs []*PtpConfigUpdate
+	files, err := ioutil.ReadDir(dir)
+	if err != nil {
+		log.Errorf("error reading all config fils %s", err)
+	}
+	for _, file := range files {
+		if checkIfPtP4lConf(file.Name()) {
+			log.Info(file.Name())
```
jzding, line 199: Is this log temporary? Printing file name without any context is hard to understand in the logs.

jzding (approved):

plugins/ptp_operator/ptp4lconf/ptp4lConfig.go
aneeshkp: this event is only when you modify(add/remove)   ptp4l.x.config file, I will removed the info ,Which is not required

plugins/ptp_operator/ptp4lconf/ptp4lConfig.go
aneeshkp: changed to  log.Infof("ptp4l config file %s",file.Name())

aneeshkp: I just cleaned it .not required

PR #64, opened 2021-09-03
Fix reconnect error

aneeshkp:
This PR fixes three items
1. On delete and create of  event of pt4l config , the sidecar was cleaning up threshold values and resetting to default 100 nano secs. 
This was briefly setting offset check on  default values.
Also on delete there was an error trying to close already closed channel.
2. when  application losses connection to amqp and then reconnects , the QDR  thread monitoring for reconnects and when trying to build the AMQ object again , it loses the Override Function that was set earlier.Thsi PR fixes to retain  those values
3. Updated SDK AMQ and  go-SDK to the latest


cmd/main.go
```diff
@@ -77,6 +78,12 @@ func main() {
 	prometheus.Unregister(collectors.NewProcessCollector(collectors.ProcessCollectorOpts{}))
 	prometheus.Unregister(collectors.NewGoCollector())
 
+	nodeIP := os.Getenv("NODE_IP")
+	if nodeIP != "" {
+		amqpHost = strings.Replace(amqpHost, "NODE_IP", nodeIP, 1)
```
jzding, line 83: amqpHost is default to `amqp:localhost:5672`. Should the `NODE_IP` be `localhost` here?
```
amqpHost = strings.Replace(amqpHost, "localhost", nodeIP, 1)
```

aneeshkp: this is  an option to use NODE_IP if required , the user will  set the transport host  value as amqp://NODE_IP:30005 and  the container will read  from env  replace NODE_IP to the actual node ip value 
```
  - name: NODE_IP
              valueFrom:
                fieldRef:
                  fieldPath: status.hostIP
```

aneeshkp: only replaces the  string NODE_IP with  the ip address

jzding: got it.

jzding (approved):

PR #85, opened 2021-12-10
e2e test for verifying producer ,consumer and router

aneeshkp:
Added integration test for validating producer and consumer event

### DIFF
--- a/examples/manifests/consumer.yaml
+++ b/examples/manifests/consumer.yaml
@@ -15,15 +15,8 @@ spec:
       labels:
         app: consumer
     spec:
-      affinity:
-        nodeAffinity:
-          requiredDuringSchedulingIgnoredDuringExecution:
-            nodeSelectorTerms:
-              - matchExpressions:
-                  - key: app
-                    operator: In
-                    values:
-                      - local
+      nodeSelector:
+        node-role.kubernetes.io/worker: ""
       serviceAccountName: sidecar-consumer-sa
       containers:
         - name: cloud-native-event-consumer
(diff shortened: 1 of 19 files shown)


examples/manifests/consumer.yaml
```diff
@@ -15,15 +15,8 @@ spec:
...
-        nodeAffinity:
-          requiredDuringSchedulingIgnoredDuringExecution:
-            nodeSelectorTerms:
-              - matchExpressions:
-                  - key: app
-                    operator: In
-                    values:
-                      - local
+      nodeSelector:
+        node-role.kubernetes.io/worker: ""
```
jzding, line 19: This might not work in CI. There are multiple worker nodes in openshift aws CI environment. In hw-event-proxy e2e I kept the above affinity code and marked the first `Ready` worker node as `app=local`.
```
# label the first Ready worker node as local
label-node:
	@kubectl label --overwrite node $(shell kubectl get nodes -l node-role.kubernetes.io/worker="" | grep Ready | cut -f1 -d" " | head -1) app=local
```

jzding (commented):
Please update ginkgo version in hack/run-functests.sh:
```
	go get github.com/onsi/ginkgo/ginkgo@v1.16.5
```

examples/manifests/consumer.yaml
aneeshkp: What I am doing here is not trying to land all under the same worker node. for CI it doesn't matter.

aneeshkp: Alright added affinity back

jzding (approved):

PR #91, opened 2022-01-27
Add Dockerfile.dev for quay.io build triggers

jzding:

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -1,0 +1,24 @@
+# Used by quay.io to trigger developer builds
+FROM quay.io/redhat-cne/openshift-origin-release:golang-1.17 AS builder
+
+ENV GO111MODULE=off
+ENV CGO_ENABLED=1
+ENV COMMON_GO_ARGS=-race
+ENV GOOS=linux
+ENV GOPATH=/go
+
+WORKDIR /go/src/github.com/redhat-cne/cloud-event-proxy
+COPY . .
+
+RUN hack/build-go.sh
+
+FROM quay.io/redhat-cne/openshift-origin-release:base-4.10 AS bin
+COPY --from=builder /go/src/github.com/redhat-cne/cloud-event-proxy/build/cloud-event-proxy /
+COPY --from=builder /go/src/github.com/redhat-cne/cloud-event-proxy/plugins/*.so /plugins/
+
+LABEL io.k8s.display-name="Cloud Event Proxy" \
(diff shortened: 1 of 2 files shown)


examples/consumer.Dockerfile.dev
```diff
@@ -0,0 +1,21 @@
+# Used by quay.io to trigger developer builds
+FROM quay.io/redhat-cne/openshift-origin-release:golang-1.17 AS builder
```
aneeshkp, line 2: is this a good idea to keep a copy of openshift-origin-release here ?

Dockerfile.dev
```diff
@@ -0,0 +1,24 @@
+# Used by quay.io to trigger developer builds
+FROM quay.io/redhat-cne/openshift-origin-release:golang-1.17 AS builder
```
aneeshkp, line 2: I am not sure if this is okay , do we have a some kind of process  to follow  openshift release and update this ?

aneeshkp: I guess it is okay, just wanted to make sure this is not used downstream. because brew doesn't have access to pulling from external registries

nishant-parekh: so from a downstream builds perspective I think we are going to continue using Dockerfile and this Dockerfile.dev is used for development and upstream testing. Is that correct @jzding ?

jzding: yes. I expect Dockerfile.dev is used only for developer tests. I put the comment at the the first line that it is for `developer builds`. Downstream should use Dockerfile and official `origin-` images under quay.io/openshift.

Dockerfile.dev
```diff
@@ -0,0 +1,24 @@
...
+FROM quay.io/redhat-cne/openshift-origin-release:base-4.10 AS bin
+COPY --from=builder /go/src/github.com/redhat-cne/cloud-event-proxy/build/cloud-event-proxy /
+COPY --from=builder /go/src/github.com/redhat-cne/cloud-event-proxy/plugins/*.so /plugins/
+
+LABEL io.k8s.display-name="Cloud Event Proxy" \
+      io.k8s.description="This is a component of OpenShift Container Platform and provides a side car to handle cloud events." \
+      io.openshift.tags="openshift" \
+      maintainer="Aneesh Puttur <aputtur@redhat.com>"
+
+ENTRYPOINT ["./cloud-event-proxy"]
```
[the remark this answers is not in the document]
jzding: added

nishant-parekh (approved):

PR #98, opened 2022-02-15
CNF-4018 Enable Flexible Registration for Multiple PTP Events clock-class-change,status and os-sync

aneeshkp:
Enable Flexible Registration for Multiple PTP Events
This adds the ability to subscribe to multiple events.
For each Publisher/Subscriber you create separate AMQ sender/listener objects

### DIFF
--- a/examples/consumer/main.go
+++ b/examples/consumer/main.go
@@ -151,17 +151,14 @@ func createSubscription(resourceAddress string) (sub pubsub.PubSub, err error) {
 	if subB, err = json.Marshal(&sub); err == nil {
 		rc := restclient.New()
 		if status, subB = rc.PostWithReturn(subURL, subB); status != http.StatusCreated {
-			err = fmt.Errorf("subscription creation api at %s, returned status %d", subURL, status)
-			return
+			err = fmt.Errorf("error subscription creation api at %s, returned status %d", subURL, status)
+		} else {
+			err = json.Unmarshal(subB, &sub)
 		}
 	} else {
-		log.Error("failed to marshal subscription ")
+		err = fmt.Errorf("failed to marshal subscription or %s", resourceAddress)
 	}
-	if err = json.Unmarshal(subB, &sub); err != nil {
-		return
-	}
-	return sub, err
-
+	return
 }
 
 func createPublisherForStatusPing(resourceAddress string) []byte {
--- a/examples/consumer/main.go
+++ b/examples/consumer/main.go
@@ -27,6 +27,8 @@ import (
 	"github.com/redhat-cne/cloud-event-proxy/pkg/common"
 	"github.com/redhat-cne/cloud-event-proxy/pkg/restclient"
 	"github.com/redhat-cne/sdk-go/pkg/event"
+	ptpEvent "github.com/redhat-cne/sdk-go/pkg/event/ptp"
+	redfishEvent "github.com/redhat-cne/sdk-go/pkg/event/redfish"
 	"github.com/redhat-cne/sdk-go/pkg/pubsub"
 	"github.com/redhat-cne/sdk-go/pkg/types"
 	v1pubsub "github.com/redhat-cne/sdk-go/v1/pubsub"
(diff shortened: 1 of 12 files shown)


examples/consumer/main.go
```diff
@@ -145,25 +136,32 @@ RETRY:
...
-		if status, b := rc.PostWithReturn(subURL, b); status == http.StatusCreated {
-			return b
+		if status, subB = rc.PostWithReturn(subURL, subB); status != http.StatusCreated {
+			err = fmt.Errorf("subscription creation api at %s, returned status %d", subURL, status)
+			return
 		}
-		log.Errorf("subscription create returned error %s", string(b))
 	} else {
-		log.Errorf("failed to create subscription ")
+		log.Error("failed to marshal subscription ")
```
jzding, line 158: add return here.

aneeshkp: done

jzding:
look good to me. I will let Nishant to add the label so that he has a chance to review.

plugins/mock/mock_plugin.go
```diff
@@ -113,19 +113,19 @@ func createMockEvent(pub pubsub.PubSub) (ceEvent.Event, error) {
 	data := ceEvent.Data{
 		Version: "v1",
 		Values: []ceEvent.DataValue{{
-			Resource:  string(ptp.SyncStatusState),
+			Resource:  "/mock",
 			DataType:  ceEvent.NOTIFICATION,
 			ValueType: ceEvent.ENUMERATION,
 			Value:     ptp.ACQUIRING_SYNC,
```
nishant-parekh, line 119: nit: may be want remove any reference to ptp-event in mock_plugin. can we set this to an arbitary string

hack/deploy_test.sh
```diff
@@ -24,17 +25,27 @@ if [ "$action" == "undeploy" ]; then
...
  deploy_consumer $action $NamespaceConsumerTesting $NamespaceAMQTesting || true
  deploy_producer $action $NamespaceProducerTesting $NamespaceAMQTesting || true
-
-fi
-
-
-
-
-
+fi
```
nishant-parekh, line 51: add an extra carriage return to avoid github warnings

plugins/ptp_operator/ptp_operator_plugin.go
```diff
@@ -70,17 +70,25 @@ func Start(wg *sync.WaitGroup, configuration *common.SCConfiguration, fn func(e
 	config = configuration
 	// register metrics type
 	ptpMetrics.RegisterMetrics(nodeName)
+	publishers = InitPubSubTypes()
 
 	// 1. Create event Publication
-	var pub pubsub.PubSub
+	// 1. Create event Publication
```
nishant-parekh, line 76: remove

plugins/ptp_operator/ptp_operator_plugin.go
```diff
@@ -314,3 +325,21 @@ func HasEqualInterface(a []*string, b []*ptp4lconf.PTPInterface) bool {
...
+		Resource:  ptp.OsClockSyncState,
+	}
+	publishers[ptp.PtpClockClassChange] = &ptpTypes.EventPublisherType{
+		EventType: ptp.PtpClockClassChange,
+		Resource:  ptp.PtpClockClass,
+	}
+	publishers[ptp.PtpStateChange] = &ptpTypes.EventPublisherType{
+		EventType: ptp.PtpStateChange,
+		Resource:  ptp.PtpLockState,
+	}
```
nishant-parekh, line 343: There are a few other event Types defined in the ptp event types file do we need to add the publishers for those eventTypes too?
https://github.com/redhat-cne/sdk-go/blob/main/pkg/event/ptp/types.go

plugins/ptp_operator/ptp_operator_plugin.go
```diff
@@ -256,7 +264,10 @@ func Start(wg *sync.WaitGroup, configuration *common.SCConfiguration, fn func(e
 		return nil
 	}
 	log.Infof("setting up status listener")
```
nishant-parekh, line 266: update log message to: `setting up status listener for each publisher type`

nishant-parekh (changes requested):
overall looks good some small comments

plugins/mock/mock_plugin.go
aneeshkp: fixed

plugins/ptp_operator/ptp_operator_plugin.go
aneeshkp: Not yet, right now we only support 3 types of events that are implemented. until we don't know what is needed, we are not adding it. The list is from API spec document

plugins/ptp_operator/ptp_operator_plugin.go
aneeshkp: fixed

nishant-parekh (approved):

PR #103, opened 2022-03-23
Use http instead of fasthttp for consumer

jzding:
Switch to http to remove the `Path traversal` security warnings
introduced by order fasthttp package.
The http server also supports ipv6, while fasthttp does not.


aneeshkp (approved):
/lgtm

PR #104, opened 2022-03-24
Bug 2067312: PPT event source is lost when received by the consumer

aneeshkp:
Fixes lost event source
Also minor fix on  ping status call returning 404 error
plus nil pointer exception when PTP starts with the faulty interface.

### DIFF
--- a/plugins/ptp_operator/metrics/metrics.go
+++ b/plugins/ptp_operator/metrics/metrics.go
@@ -460,7 +460,10 @@ func (p *PTPEventManager) ExtractMetrics(msg string) {
 				// update role metrics
 				UpdateInterfaceRoleMetrics(processName, ptpIFace, role)
 			}
-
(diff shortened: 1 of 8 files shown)


plugins/ptp_operator/metrics/metrics.go
```diff
@@ -460,7 +460,10 @@ func (p *PTPEventManager) ExtractMetrics(msg string) {
 				// update role metrics
 				UpdateInterfaceRoleMetrics(processName, ptpIFace, role)
 			}
-
+			if _, ok := ptpStats[master]; !ok { //
+				log.Errorf("no offset stats found for master for  portid %d with role %s (the port started in fault state ", portID, role)
```
jzding, line 464: nit: (the port started in fault state`)` <= closing bracket

jzding (approved):

PR #106, opened 2022-03-25
Bug 2067311: PTP event source is lost when received by the consumer 

aneeshkp:
Fixes lost event source
plus nil pointer exception when PTP starts with the faulty interface.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,8 +9,8 @@ require (
 	github.com/onsi/ginkgo v1.16.5
 	github.com/onsi/gomega v1.17.0
 	github.com/prometheus/client_golang v1.11.0
-	github.com/redhat-cne/rest-api v0.1.1-0.20220105171308-cc511366a079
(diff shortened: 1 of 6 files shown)


jzding:
`Also minor fix on ping status call returning 404 error` - do you mean the change in [plugins/ptp_operator/ptp_operator_plugin.go](https://github.com/redhat-cne/cloud-event-proxy/pull/104/files#diff-5b986712d51387b70a04b8619f1fba5bb1c4b11897f71e571f16190d4e500b24)? I don't see it here.

aneeshkp:
> `Also minor fix on ping status call returning 404 error` - do you mean the change in [plugins/ptp_operator/ptp_operator_plugin.go](https://github.com/redhat-cne/cloud-event-proxy/pull/104/files#diff-5b986712d51387b70a04b8619f1fba5bb1c4b11897f71e571f16190d4e500b24)? I don't see it here. 

"No, there are multiple event types in 4.10 " - I fixed the description. Good catch

jzding (approved):

PR #111, opened 2022-04-12
Generate ptp4l/phc2sys process status metrics and events

aneeshkp:
linux ptpdaemon will be logging no of times ptp4l/phc2sys process dies and restarts.
The event framewortk will parse the logs for these messages and generate metrics
 to capture process restart count and current status of the ptp4l/phc2sys process
Also send events when ptp4l/phc2sys is down, notificying subscribers on status of the ptp clock.

### DIFF
--- a/plugins/ptp_operator/metrics/registry.go
+++ b/plugins/ptp_operator/metrics/registry.go
@@ -82,6 +82,24 @@ var (
 			Name:      "clock_class",
 			Help:      "",
 		}, []string{"process", "node"})
+
+	// ProcessStatus  ... update process status
+	ProcessStatus = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: ptpNamespace,
+			Subsystem: ptpSubsystem,
+			Name:      "process_status",
+			Help:      "0 = DOWN, 1 = UP",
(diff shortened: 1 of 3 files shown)


plugins/ptp_operator/metrics/registry.go
```diff
@@ -171,3 +191,13 @@ func DeleteInterfaceRoleMetrics(process, ptpInterface string) {
 	InterfaceRole.Delete(prometheus.Labels{
 		"process": process, "node": ptpNodeName, "iface": ptpInterface})
 }
+
+// UpdateProcessStatusMetrics  -- update process status metrics
+func UpdateProcessStatusMetrics(process, cfgName string, status int64) {
+	ProcessStatus.With(prometheus.Labels{
+		"process": process, "node": ptpNodeName, "config": cfgName}).Set(float64(status))
+	if status == PtpProcessUp {
```
jzding, line 199: Since there is no caching of previous status, if the process stays up and we keep checking the logs, this `restart count` will keep increasing even though there is no restart. The logic does not check out.

aneeshkp: The logs are read in real-time. We don't go back in time, just read the present time.

jzding (approved):

PR #123, opened 2022-06-15
Bug 2091599 :event is sent multiple times when non slave port goes to faulty.

aneeshkp:
The fix makes sure events are not triggered when master port goes to faulty state.


aneeshkp:
/retitle Bug 2091599: event is sent multiple times when non slave port goes to faulty.

PR #133, opened 2022-08-18
Http transport 4.12

aneeshkp:
Add http support to cloud event proxy


cmd/main.go
```diff
@@ -44,27 +45,31 @@ import (
...
+	pluginHandler      plugins.Handler
 )
 
 func main() {
 	// init
 	common.InitLogger()
 	flag.StringVar(&metricsAddr, "metrics-addr", ":9091", "The address the metric endpoint binds to.")
 	flag.StringVar(&storePath, "store-path", ".", "The path to store publisher and subscription info.")
-	flag.StringVar(&amqpHost, "transport-host", "amqp:localhost:5672", "The transport bus hostname or service name.")
+	flag.StringVar(&transportHost, "transport-host", "amqp:localhost:5672", "The transport bus hostname or service name.")
```
jzding, line 70: How about set default value to `localhost:9043` for local http testing?
`flag.StringVar(&transportHost, "transport-host", "localhost:9043", "The transport bus hostname or service name.")`

aneeshkp: For http it should never set to localhost, it should be k8s service url

cmd/main.go
```diff
@@ -44,27 +45,31 @@ import (
...
 	flag.IntVar(&apiPort, "api-port", 8089, "The address the rest api endpoint binds to.")
+	flag.StringVar(&httpEventPublisher, "http-event-publishers", "", "Comma separated address of the publishers available.")
```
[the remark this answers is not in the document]
aneeshkp: Same here, for http it should never set to localhost, it should be k8s service url

jzding (approved):
/lgtm

PR #134, opened 2022-08-29
Fix readme and update SDK

aneeshkp:
Fixed readme, added logic for ping status and updated SDK changes


README.md
```diff
@@ -20,6 +23,69 @@
...
+            - "--transport-host=amqp://router.router.svc.cluster.local"
+            - "--api-port=8089"
+```
+#### HTTP Protocol
+##### Producer
+CloudEvents HTTP Protocol will be enabled based on  url in `transport-host`.
+If HTTP is identified then the publisher will start a publisher rest service, which is accessible outside the container via k8s service name.
+The Publisher service will have the ability to register  consumer endpoints to publish events.
+
+HTTP consumer example
```
jzding, line 61: HTTP producer

jzding:
Make this debug: https://github.com/redhat-cne/cloud-event-proxy/blob/main/cmd/main.go#L234

jzding (approved):

PR #135, opened 2022-09-01
Fix pingStatus

aneeshkp:
ping status fires events for all subscribers, the fix will isolate only the called client


jzding (approved):
/lgtm

PR #137, opened 2022-10-03
Implement http currentState api call as per O-Ran Spec

aneeshkp:
Added CurrentState api call to get direct response from the api reporting back the state of the  requested resource address
localhost:ip/vi/{resourceAddress}/CurrentState


examples/consumer/main.go
```diff
@@ -210,6 +215,21 @@ func pingForStatus(resource string, resourceID string) {
 	}
 }
 
+// pingForStatus sends pings to fetch events status
```
jzding, line 218: update function name

jzding (approved):

examples/consumer/main.go
aneeshkp: done

PR #138, opened 2022-10-04
typo fix on building source of the event

aneeshkp:
added missing forward slash

### DIFF
--- a/plugins/ptp_operator/metrics/manager.go
+++ b/plugins/ptp_operator/metrics/manager.go
@@ -164,7 +164,7 @@ func (p *PTPEventManager) GetPTPEventsData(state ptp.SyncState, ptpOffset int64,
 		return nil
 	}
 	// /cluster/xyz/ptp/CLOCK_REALTIME this is not address the event is published to
-	eventSource := fmt.Sprintf(p.resourcePrefix, p.nodeName, source)
+	eventSource := fmt.Sprintf(p.resourcePrefix, p.nodeName, fmt.Sprintf("/%s", source))
 	data := ceevent.Data{
 		Version: "v1",
 		Values: []ceevent.DataValue{{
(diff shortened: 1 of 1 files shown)


plugins/ptp_operator/metrics/manager.go
```diff
@@ -164,7 +164,7 @@ func (p *PTPEventManager) GetPTPEventsData(state ptp.SyncState, ptpOffset int64,
 		return nil
 	}
 	// /cluster/xyz/ptp/CLOCK_REALTIME this is not address the event is published to
-	eventSource := fmt.Sprintf(p.resourcePrefix, p.nodeName, source)
+	eventSource := fmt.Sprintf(p.resourcePrefix, p.nodeName, fmt.Sprintf("/%s", source))
```
jzding, line 167: would this work better?
```
eventSource := fmt.Sprintf(p.resourcePrefix, p.nodeName, "/", source)
```

plugins/ptp_operator/metrics/manager.go
```diff
@@ -207,7 +207,7 @@ func (p *PTPEventManager) PublishEvent(state ptp.SyncState, ptpOffset int64, eve
 		return
 	}
 	// /cluster/xyz/ptp/CLOCK_REALTIME this is not address the event is published to
-	eventSource = fmt.Sprintf(p.resourcePrefix, p.nodeName, eventSource)
+	eventSource = fmt.Sprintf(p.resourcePrefix, p.nodeName, fmt.Sprintf("/%s", eventSource))
```
jzding, line 210: same comment, also looks better if you name function arg as `source` instead of `eventSource` to avoid assign an argument to itself.

plugins/ptp_operator/metrics/manager.go
aneeshkp: resourcePrefix         = "/cluster/node/%s%s" with two parameter  only
So it won't work

plugins/ptp_operator/metrics/manager.go
aneeshkp: updated

jzding (approved):
/lgtm

PR #142, opened 2022-10-20
Amq current state API call fix 

aneeshkp:
Event Pull Status Notification GET Method not working as per o-Ran spec

### DIFF
--- a/examples/consumer.Dockerfile
+++ b/examples/consumer.Dockerfile
@@ -1,4 +1,4 @@
-FROM openshift/origin-release:golang-1.15 AS builder
+FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.19-openshift-4.12 AS builder
 ENV GO111MODULE=on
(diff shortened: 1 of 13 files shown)


examples/manifests/kustomization.yaml
```diff
@@ -14,5 +14,5 @@ images:
   newName: quay.io/redhat-cne/cloud-event-consumer
   newTag: latest
 - name: cloud-event-sidecar
-  newName: quay.io/redhat-cne/cloud-event-proxy
+  newName: quay.io/aneeshkp/origin-cloud-event-proxy
```
nishant-parekh, line 17: I think this needs to point to the redhat-cne version

nishant-parekh (commented):
small update but otherwise looks good

examples/manifests/kustomization.yaml
aneeshkp: Done

nishant-parekh (approved):

PR #144, opened 2022-10-25
Lint patch

aneeshkp:
Updated linter and fixed code as per make lint


.golangci.yaml
```diff
@@ -5,42 +5,72 @@ run:
...
 
 linters:
   disable-all: true
   enable: # NOTE: please keep this list alphabetically sorted
     - bodyclose
     - deadcode
+    - decorder
     - dogsled
-    - errcheck
+    #- errcheck
```
jzding, line 60: why disable this?

aneeshkp: it was not working before, but with this fix now turning  on that will generate many fixed where we just ignore the  error.

jzding (approved):
/lgtm

PR #147, opened 2022-11-08
[release-4.11] Bug OCPBUGS-3295: Event Pull Status Notification GET Method not working as per o-Ran spec

aneeshkp:
fix get current state for events

### DIFF
--- a/plugins/ptp_operator/ptp_operator_plugin.go
+++ b/plugins/ptp_operator/ptp_operator_plugin.go
@@ -19,7 +19,12 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"strings"
 	"sync"
+
+	"k8s.io/utils/pointer"
+
+	"github.com/redhat-cne/sdk-go/pkg/event"
 
 	"github.com/redhat-cne/cloud-event-proxy/plugins/ptp_operator/ptp4lconf"
 
@@ -33,6 +38,7 @@ import (
 	ptpSocket "github.com/redhat-cne/cloud-event-proxy/plugins/ptp_operator/socket"
 	ptpTypes "github.com/redhat-cne/cloud-event-proxy/plugins/ptp_operator/types"
 	v1amqp "github.com/redhat-cne/sdk-go/v1/amqp"
+
 	log "github.com/sirupsen/logrus"
 
 	ptpMetrics "github.com/redhat-cne/cloud-event-proxy/plugins/ptp_operator/metrics"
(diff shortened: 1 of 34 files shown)


plugins/ptp_operator/ptp_operator_plugin.go
```diff
@@ -175,6 +160,73 @@ func Start(wg *sync.WaitGroup, configuration *common.SCConfiguration, fn func(e
...
+// getCurrentStatOverrideFn is called when current state is received by rest api
+func getCurrentStatOverrideFn() func(e v2.Event, d *channel.DataChan) error {
+	return func(e v2.Event, d *channel.DataChan) error {
+		if e.Source() != "" {
+			log.Infof("setting return address to %s", e.Source())
+			d.ReturnAddress = pointer.StringPtr(e.Source())
+		}
+		log.Infof("got status check call,send events for subscriber %s => %s", d.ClientID.String(), e.Source())
+		var eventType ptp.EventType
+		if !strings.Contains(e.Source(), string(ptp.PtpLockState)) &&
```
jzding, line 172: could just add a `else` statement at line 184

examples/consumer/deployment/namespace.yaml
```diff
@@ -5,3 +5,7 @@ metadata:
   labels:
     name: cloud-events
     openshift.io/cluster-monitoring: "true"
+    security.openshift.io/scc.podSecurityLabelSync: "false"
+    pod-security.kubernetes.io/audit: "privileged"
+    pod-security.kubernetes.io/enforce: "privileged"
+    pod-security.kubernetes.io/warn: "privileged"
```
jzding, line 11: what do these do?

plugins/ptp_operator/ptp_operator_plugin.go
aneeshkp: can we turn off that  earl return linter error?

plugins/ptp_operator/ptp_operator_plugin.go
```diff
@@ -175,6 +160,73 @@ func Start(wg *sync.WaitGroup, configuration *common.SCConfiguration, fn func(e
...
+			!strings.Contains(e.Source(), string(ptp.OsClockSyncState)) &&
+			!strings.Contains(e.Source(), string(ptp.OsClockSyncState)) {
```
jzding, line 174: I prefer to not disable and lint checks. If we keep this then need to fix this line. I believe it should check for `ptp.PtpClockClass` instead.

aneeshkp: good catch, Fixed that now

PR #163, opened 2022-12-15
Bug OCPBUGS-4953: PTP events - error message on even log needs to be remove

aneeshkp:
removed error from log when parsing invalid no of fields 


jzding (approved):
/lgtm

PR #176, opened 2023-01-03
[release-4.10] OCPBUGS-4755: update sdk-go model for increasing amq timeout

jzding:


aneeshkp (changes requested):
The log wait for 7 minutes is preventing linuxptpdaemon throw an error continuously and wait for 7 minutes.

PR #192, opened 2023-02-03
[release-4.10] Bug OCPBUGS-4755: Amq connection configurable timeout and nohup

jzding:
Allow configurable timeout for initial amq connection.

Introduce amqp://nohup transportHost for local logging of events when amq connection is not available


aneeshkp (approved):
/lgtm

PR #194, opened 2023-02-10
Bug OCPBUGS-2832: update sdk-go for http restart fix

jzding:


go.mod
```diff
@@ -10,8 +10,8 @@ require (
 	github.com/onsi/ginkgo v1.16.5
 	github.com/onsi/gomega v1.22.1
 	github.com/prometheus/client_golang v1.13.0
-	github.com/redhat-cne/rest-api v0.1.1-0.20230119134643-a59307697ba2
-	github.com/redhat-cne/sdk-go v0.1.1-0.20230119111541-38e32b882917
+	github.com/redhat-cne/rest-api v0.1.1-0.20230210141139-50f5dd837cf8
+	github.com/redhat-cne/sdk-go v1.5.0
```
nishant-parekh, line 14: This change I added to try and track the updates for sdk-go.. the idea is the we update minor version per release:
1.1 => 4.9
1.2 => 4.10
1.3 => 4.11
1.4 => 4.12
1.5 => 4.13

Do we want to use this approach going forward? else I would suggest I will delete this release so it can revert back to 0.1.1-XXX format release it used to so far..

nishant-parekh (commented):
@jzding @aneeshkp this change pulls some of the releases I created to streamline our updates for sdk-go. If this looks good we can replicate that is rest-api.. if not I can revert my releases.. and we can update this PR.. so Jack's changes can go through.

go.mod
jzding: updated after reversing the tags.

nishant-parekh (approved):

PR #195, opened 2023-02-10
Add multi node support for http

aneeshkp:
Add multi node support for HTTP .
 This needs to be back ported to 4.12

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -21,7 +21,6 @@ import (
 	"os"
 	"os/signal"
 	"path"
-	"strings"
 	"sync"
 	"syscall"
 	"time"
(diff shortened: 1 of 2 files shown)


cmd/main.go
```diff
@@ -86,21 +85,7 @@ func main() {
 
 	nodeIP := os.Getenv("NODE_IP")
 	nodeName := os.Getenv("NODE_NAME")
```
jzding, line 87: move these 2 lines inside the SanitizeTransportHost() function so that we do not need the extra 2 arguments.
```
transportHost = common.SanitizeTransportHost(transportHost)
```

PR #205, opened 2023-02-21
BUG OCPBUGS-7826: Add multi node support for consumer to identify publishers

aneeshkp:
when deploying consumer app, if NODE_NAME is used as placeholder for publisher URL , the sidecar was skipping replacement and this was resulting from identifying valid .publisher service endpoint.

### DIFF
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -365,7 +365,6 @@ func APIHealthCheck(uri *types.URI, delay time.Duration) (ok bool, err error) {
 
 // HTTPTransportHealthCheck ... http transport should be ready before starting to consume events
 func HTTPTransportHealthCheck(uri *types.URI, delay time.Duration) (ok bool, err error) {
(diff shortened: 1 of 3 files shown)


pkg/common/common.go
```diff
@@ -365,7 +365,7 @@ func APIHealthCheck(uri *types.URI, delay time.Duration) (ok bool, err error) {
 
 // HTTPTransportHealthCheck ... http transport should be ready before starting to consume events
 func HTTPTransportHealthCheck(uri *types.URI, delay time.Duration) (ok bool, err error) {
-	log.Printf("checking for http transport health\n")
+	log.Printf("HTTPTransportHealthCheckn")
```
jzding, line 368: Checkn -> Check\n

aneeshkp: Done

jzding (approved):
/lgtm

PR #211, opened 2023-02-28
OCPBUGS-7811: remove ptp process metrics when ptpcfig is  deleted

aneeshkp:
This fixes retaining ptp process metrics when ptpconfig is deleted
OCPBUGS-7811 WPC GM - a new process file is created when applying a deleted gm ptpconfig


jzding (approved):
/lgtm

PR #216, opened 2023-03-02
Add example to use event subscription without using side car

aneeshkp:
Shows examples of how to HTTP protocol directly with the producer.
This is not tested or supported yet

### DIFF
--- a/examples/simplehttp/common/httpclient.go
+++ b/examples/simplehttp/common/httpclient.go
@@ -1,0 +1,62 @@
+package common
+
+import (
+	"context"
+	"errors"
+	"io"
+	"net/http"
+	"syscall"
+	"time"
+
+	cloudevents "github.com/cloudevents/sdk-go/v2"
+	log "github.com/sirupsen/logrus"
+)
+
+var (
+	cancelTimeout = 2 * time.Second
+)
+
+// Post ...
+func Post(address string, e cloudevents.Event) (int, error) {
+	sendCtx, sendCancel := context.WithTimeout(context.Background(), cancelTimeout)
+	defer sendCancel()
+	p, err := cloudevents.NewHTTP()
+	if err != nil {
+		log.Errorf("failed to create protocol: %s", err.Error())
+		return http.StatusBadRequest, err
+	}
+	c, err := cloudevents.NewClient(p, cloudevents.WithUUIDs(), cloudevents.WithTimeNow())
+	if err != nil {
+		log.Errorf("failed to create http client: %s", err.Error())
+		return http.StatusBadRequest, err
+	}
+	log.Infof("posting now %s, to  %s", e.Type(), address)
+	e.SetDataContentType(cloudevents.ApplicationJSON)
+	ctx := cloudevents.ContextWithTarget(sendCtx, address)
+	result := c.Send(ctx, e)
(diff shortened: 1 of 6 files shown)


examples/simplehttp/common/httpclient.go
```diff
@@ -0,0 +1,69 @@
...
+// Post ...
+func Post(address string, e cloudevents.Event) (int, error) {
+	sendCtx, sendCancel := context.WithTimeout(context.Background(), cancelTimeout)
+	defer sendCancel()
+	p, err := cloudevents.NewHTTP()
+	if err != nil {
+		log.Errorf("failed to create protocol: %s", err.Error())
+		return http.StatusBadRequest, err
+	}
+	c, err := cloudevents.NewClient(p, cloudevents.WithUUIDs(), cloudevents.WithTimeNow())
```
jzding, line 28: Is this c=NewClient object cleaned up somewhere?

examples/simplehttp/common/httpclient.go
```diff
@@ -0,0 +1,69 @@
...
+	if response.StatusCode == http.StatusOK {
+		bodyBytes, err = io.ReadAll(response.Body)
+		if err != nil {
+			return []byte(err.Error()), http.StatusBadRequest, err
+		}
+	} else {
+		bodyBytes, err = io.ReadAll(response.Body)
+		if err != nil {
+			return []byte(err.Error()), http.StatusBadRequest, err
+		}
```
jzding, line 66: Same actions are performed for if{} and else{}, then there is no need for if-else statement.

examples/simplehttp/main.go
```diff
@@ -0,0 +1,184 @@
...
+
+func help() {
+	fmt.Println("--------------------------------")
+	fmt.Println("1. setting up port forwarding to the container port to expose the service external to k8s")
+	fmt.Println(" oc get pods -n openshift-ptp -l app=linuxptp-daemon")
+	fmt.Println("NAME                            READY   STATUS    RESTARTS   AGE")
+	fmt.Println("linuxptp-daemon-xmjcr           3/3     Running   0          13h")
+	fmt.Println("")
+	fmt.Println("Run following command to set port forwarding")
+	fmt.Println("oc -n openshift-ptp linuxptp-daemon-xmjcr 9043:9043 --address='0.0.0.0'")
```
jzding, line 161: not a correct command

examples/simplehttp/common/httpclient.go
aneeshkp: provided transport object and func scoped

examples/simplehttp/common/httpclient.go
aneeshkp: good one

examples/simplehttp/external-service.yaml
```diff
@@ -0,0 +1,24 @@
...
+      targetPort: 3306
+      nodePort: 0
+---
+apiVersion: "v1"
+kind: "Endpoints"
+metadata:
+  name: "event-consumer-external"
+subsets:
+  - addresses:
+      - ip: "10.22.32.248"  #change this to your laptop ip address
```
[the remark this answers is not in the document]
aneeshkp: 10.x is private. wont matter i guess

jzding (approved):
/lgtm

PR #217, opened 2023-03-02
OCPBUGS-7974: PTP BC: an unnecessary openshift_ptp_clock_state metric is shown

aneeshkp:
Stop reporting wrong metrics


jzding (approved):
/lgtm

PR #223, opened 2023-03-10
OCPBUGS-8714: With GM config, phc2sys clock lock state is shown as FREERUN in ptp metrics while it should be LOCKED

aneeshkp:
Fix CLOCK_REALTIME not updated in metrics with ts2phc enabled

### DIFF
--- a/plugins/ptp_operator/metrics/metrics.go
+++ b/plugins/ptp_operator/metrics/metrics.go
@@ -190,11 +190,13 @@ func (p *PTPEventManager) ExtractMetrics(msg string) {
 		}
 		// if it was ts2phc then it was considered as master offset
 		interfaceType := types.IFace(master)
-		if masterOffsetSource == ts2phcProcessName {
+		if processName == ts2phcProcessName { // if current offset is read from ts2phc
+			// ts2phc return actual interface name unlike ptp4l
 			ptpInterface = ptp4lconf.PTPInterface{Name: interfaceName}
 		} else {
 			// for ts2phc there is no slave interface configuration
-			interfaceType = types.IFace(interfaceName)
+			// fort pt4l find the slave configured
+			interfaceType = types.IFace(interfaceName) // this will be CLOCK_REALTIME or master
 			ptpInterface, _ = ptp4lCfg.ByRole(types.SLAVE)
 		}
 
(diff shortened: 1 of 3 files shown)


plugins/ptp_operator/metrics/metrics.go
```diff
@@ -210,7 +212,10 @@ func (p *PTPEventManager) ExtractMetrics(msg string) {
 		switch interfaceName {
 		case ClockRealTime: // CLOCK_REALTIME is active slave interface
 			// copy  ClockRealTime value to current slave interface
-			if r, ok := ptpStats[master]; ok && r.Role() == types.SLAVE { // publish event only if the master role is active
+			if masterOffsetSource == ts2phcProcessName {
+				//TODO: once ts2phc events are identified we need to add that here
+				p.GenPTPEvent(profileName, ptpStats[interfaceType], interfaceName, int64(ptpOffset), syncState, eventType)
```
nishant-parekh, line 217: will this be removed when we finish the above TODO or do we send both events in the `ts2phc` case

plugins/ptp_operator/metrics/ptp4lParse.go
```diff
@@ -64,15 +64,15 @@ func (p *PTPEventManager) ParsePTP4l(processName, configName, profileName, outpu
...
+		if role == types.SLAVE || masterOffsetSource == ts2phcProcessName {
 			// initialize
 			if _, found := ptpStats[ClockRealTime]; !found {
 				ptpStats[ClockRealTime] = stats.NewStats(configName)
 				ptpStats[ClockRealTime].SetProcessName(phc2sysProcessName)
 			}
 			if _, found := ptpStats[master]; !found {
 				ptpStats[master] = stats.NewStats(configName)
-				ptpStats[master].SetProcessName(ptp4lProcessName)
+				ptpStats[master].SetProcessName(masterOffsetSource)
```
nishant-parekh, line 75: will the masterOffsetSource be set to the correct value if we reach here for the `ptp4l`case?

plugins/ptp_operator/metrics/metrics.go
aneeshkp: Yes, we are not handling GM events now and we don't know what the GM events are going to look like.

plugins/ptp_operator/metrics/ptp4lParse.go
aneeshkp: At this point, we assume that  the ptpconfig has been configured properly. In that case, we can either have ptp4l offset or ts2phcs offset reporting. we can't have both. So when offset is read for the first time from master this will be updated

nishant-parekh (approved):

PR #232, opened 2023-03-31
Cr as store

aneeshkp:
This PR introduces Custom Resource to store event Subscribers
Cloud-event-proxy will now have k8s awareness and will perform the call to k8s API to retrieve and update Custom Resource.


aneeshkp:
closing this since configmap solution is merged

PR #235, opened 2023-04-06
Update consumer deployment and README

jzding:
Consumer deployment should use  examples/manifests/ .
Remove the obsolete consumer manifests from examples/consumer/deployment.


examples/manifests/consumer.yaml
```diff
@@ -24,7 +24,7 @@ spec:
           args:
             - "--local-api-addr=127.0.0.1:9089"
             - "--api-path=/api/ocloudNotifications/v1/"
-            - "--api-addr=127.0.0.1:8089"
```
aneeshkp, line 27: why change the port?

jzding: to be consistent with documentation and the port used at ptp/bmer operators.

README.md
```diff
@@ -73,12 +73,10 @@ HTTP producer example
           args:
             - "--metrics-addr=127.0.0.1:9091"
             - "--store-path=/store"
-            - "--ptp-event-publisher-service-NODE_NAME.openshift-ptp.svc.cluster.local:9043"
+            - "--transport-host=http://ptp-event-publisher-service-NODE_NAME.openshift-ptp.svc.cluster.local:9043
             - "--api-port=9085"
```
aneeshkp, line 77: this should be 8089 too right ?

jzding: this is producer

nishant-parekh (approved):

PR #239, opened 2023-04-17
OCPBUGS-11137: Add configmap to store subscriber info as a storage type - To prevent loss of subscribers

aneeshkp:
HTTP Publisher needs to persist subscriber in case or pod restarts the subscriber informations are not lost.
This requires permission to create configmap 
configmap are named per node. So it is necessary to set namespace and codename in the environment variable.
Consumer doesn't require any persistence of publisher  info


go.mod
```diff
@@ -74,16 +72,9 @@ require (
...
-	sigs.k8s.io/structured-merge-diff/v4 v4.2.1 // indirect
+	k8s.io/klog/v2 v2.80.1 // indirect
+	k8s.io/kube-openapi v0.0.0-20221012153701-172d655c2280 // indirect
+	sigs.k8s.io/json v0.0.0-20220713155537-f223a00ba0e2 // indirect
+	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
-
-// Manually pinned to kubernetes-1.24.0
-replace (
```
nishant-parekh, line 85: should this be deleted?

hack/test-go.sh
```diff
@@ -0,0 +1,25 @@
...
+echo "Running golang staticcheck ..."
+${BASEDIR}/bin/staticcheck --tags=test ./...
+
+echo "Running go tests..."
+KUBEBUILDER_ASSETS="$(pwd)/bin" ${GO} test \
+    --tags=test \
+    -v \
+    -covermode=count \
+    -coverprofile=coverage.out \
+    $(${GO} list ./... | grep -v e2e | tr "\n" " ")
```
nishant-parekh, line 25: missing new line

nishant-parekh (commented):
added some suggestions

examples/consumer.Dockerfile
```diff
@@ -9,7 +9,7 @@ COPY . .
 
 RUN hack/build-example-go.sh
 
-FROM openshift/origin-base AS bin
+FROM registry.ci.openshift.org/ocp/4.13:base AS bin
```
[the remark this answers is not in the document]
aneeshkp: that's consumer

Makefile
```diff
@@ -124,3 +124,10 @@ fmt-code: ## Run go fmt against code.
 
 vet: ## Run go vet against code.
 	go vet ./...
+
+generate-api:
+	hack/verify-codegen.sh
```
nishant-parekh, line 129: the script this is running doesnt exist, same for install-tools below.. should these be cleaned up as well?

nishant-parekh (commented):
looks good just one small comment

nishant-parekh (approved):

PR #250, opened 2023-05-11
fix for multinode support

edcdavid:


cmd/main.go
```diff
@@ -96,6 +97,12 @@ func main() {
 	nodeName = os.Getenv("NODE_NAME")
 	namespace = os.Getenv("NAME_SPACE")
 
+	// using the first component of the node name before the first dot (master2.example.com -> master2)
+	if nodeName != "" && strings.Contains(nodeName, ".") {
```
aneeshkp, line 101: remove this , not required. Its done in 
common.SanitizeTransportHost

plugins/ptp_operator/ptp_operator_plugin.go
```diff
@@ -73,6 +73,13 @@ var (
 func Start(wg *sync.WaitGroup, configuration *common.SCConfiguration, fn func(e interface{}) error) error { //nolint:deadcode,unused
 	// The name of NodePtpDevice CR for this node is equal to the node name
 	nodeName := os.Getenv("NODE_NAME")
+
+	// using the first component of the node name before the first dot (master2.example.com -> master2)
+	if nodeName != "" && strings.Contains(nodeName, ".") {
```
aneeshkp, line 78: same here this is not required the config that has been passed doing the same

examples/consumer/main.go
```diff
@@ -71,6 +72,12 @@ func main() {
 	flag.Parse()
 
 	nodeName := os.Getenv("NODE_NAME")
```
aneeshkp, line 74: the consumer doesn't have to do sanitize, the node name. Because  node  replacement for URI  is done by the cloud-event-proxy framework

edcdavid:
code is correct after all, removing PR

PR #253, opened 2023-07-11
fix lint errors

vitus133:
This commit fixes `golangci-lint run` errors
happening due to unused function parameters

/cc @aneeshkp @nishant-parekh @jzding


vitus133:
Linting fails on timeout

PR #255, opened 2023-07-19
[release-4.13] Bug OCPBUGS-16391: Add configmap to store subscriber info as a storage type

jzding:
Add configmap to store subscriber info as a storage type - To prevent loss of subscribers


aneeshkp (approved):
/lgtm

PR #259, opened 2023-08-03
OCPBUGS-17322: linuxptp-daemon pod status changes to Error and CrashLoopBackOff after a ptpconfig is loaded

aneeshkp:
Fixed nil pointer check


jzding (approved):
/lgtm

PR #270, opened 2023-11-30
OCPBUGS-24595: Cloudeventproxy 2 Card WPC not able to handle multiple dependency and multiple ts2phcs 

aneeshkp:


plugins/ptp_operator/event/event.go
```diff
@@ -66,14 +69,43 @@ type PMetric struct {
...
+func (dt *DependingClockState) hasClockState(iface string) (int, *ClockState) {
+	for index, dep := range *dt {
+		if *dep.IFace == iface {
+			return index, dep
+		}
+	}
+	return -1, nil
+}
+
+func (dt *DependingClockState) hasMetric() map[string]*PMetric {
```
jzding, line 81: shouldn't this return Metric based on specific iface similar to `hasClockState`?

plugins/ptp_operator/event/event.go
```diff
@@ -66,14 +69,43 @@ type PMetric struct {
...
+	for _, dep := range *dt {
+		return dep.Metric
+	}
+	return nil
+}
+func (dt *DependingClockState) hasMetricHelp() map[string]string {
```
jzding, line 87: same here, should return MetricHelp based on specific iface similar to `hasClockState`

plugins/ptp_operator/event/event.go
```diff
@@ -135,22 +176,24 @@ func (p *PTPEventState) UpdateCurrentEventState(c ClockState) ptp.SyncState {
...
-			if v.State == ptp.HOLDOVER {
-				currentState = v.State
-				break
+		for _, dd := range v {
+			if dd.State != ptp.LOCKED {
+				if dd.State == ptp.HOLDOVER {
+					currentState = dd.State
+					break
+				}
+				currentState = dd.State
```
jzding, line 193: if v[0]=FREERUN and v[1]=HOLDOVER,  currentState will be set to HOLDOVER, which is not correct

plugins/ptp_operator/event/event.go
```diff
@@ -135,22 +176,24 @@ func (p *PTPEventState) UpdateCurrentEventState(c ClockState) ptp.SyncState {
 				"node": clockState.NodeName, "iface": alias}).Set(float64(v))
 		}
 		clockState.Metric = metrics
-		p.DependsOn[c.Process] = clockState
+		p.DependsOn[c.Process] = []*ClockState{clockState}
 	}
 
 	// if all locked then its locked
 	// if anyone HOLDOVER then holdover
 	// if  anyone FREERUN then freerun
```
jzding, line 184: comments not clear. What the currentState should be if DependsOn states are a mix of HOLDVER and FREERUN?

aneeshkp: if anyone is HOLDOVER then holdover comment 2

plugins/ptp_operator/event/event.go
aneeshkp: it is based on dependent type, metrics is common for all interface based on type,  so you are checking is there is metrics object already define. You can't register same metrics twice

plugins/ptp_operator/event/event.go
aneeshkp: it is based on dependent type, metrics is common for all interface based on type,  so you are checking is there is metrics object already define. You can't register same metrics twice

jzding (approved):
/lgtm

plugins/ptp_operator/metrics/logparser.go
```diff
@@ -486,32 +494,29 @@ func (p *PTPEventManager) ParseGNSSLogs(processName, configName, output string,
...
 		if lastState != GetSyncState(syncState) || errState != nil {
 			log.Infof("%s last state %s and current state %s", processName, lastState, GetSyncState(syncState))
-			var masterAlias string
-			if m, ok := ptpStats[master]; ok {
-				masterAlias = m.Alias()
-			}
-			if alias == "" {
-				masterAlias, _ = ptp4lCfg.GetUnknownAlias()
+			masterResource := fmt.Sprintf("%s/%s", alias, MasterClockType)
+			if !p.mock { /// to enable unit testing
```
jzding, line 517: remove this, the unit test is actually checking if the event is sent.

aneeshkp: fixed it,

jzding (approved):
/lgtm

PR #272, opened 2023-11-30
[release 4.14] OCPBUGS-24302: GNSS status metrics is not showing correct state in cloud event proxy 

aneeshkp:


plugins/ptp_operator/metrics/logparser.go
```diff
@@ -437,7 +437,7 @@ func (p *PTPEventManager) ParseGNSSLogs(processName, configName, output string,
 			Offset:      pointer.Float64(gnssOffset),
 			Process:     processName,
 			IFace:       iface,
-			Value:       map[string]int64{"gnss_status": 5},
+			Value:       map[string]int64{"gnss_status": gnssState},
```
jzding, line 440: remove `else` at line 424, similar to what's in the master branch

PR #281, opened 2024-01-03
OCPBUGS-25547: PTP metrics still includes metrics for old config after deleting a ptpconfig 

aneeshkp:
fix metrics clean up for GM

### DIFF
--- a/plugins/ptp_operator/event/event.go
+++ b/plugins/ptp_operator/event/event.go
@@ -15,6 +15,7 @@
 package event
 
 import (
+	"strings"
 	"sync"
 
 	"github.com/prometheus/client_golang/prometheus"
@@ -36,6 +37,8 @@ const (
 	GNSS ClockSourceType = "gnss"
 	// DPLL ... ClockSourceType = "dpll"
 	DPLL ClockSourceType = "dpll"
+	// GM ... ClockSourceType
+	GM ClockSourceType = "GM"
 )
 
 // DependingClockState ...
@@ -45,7 +48,7 @@ type DependingClockState []*ClockState
 type PTPEventState struct {
 	sync.Mutex
 	CurrentPTPStateEvent ptp.SyncState
-	DependsOn            map[string]DependingClockState // [dpll][ens2f0]State
+	DependsOn            map[string]DependingClockState // [dpll][]*ClockState
 	Type                 ptp.EventType
 }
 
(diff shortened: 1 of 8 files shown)


plugins/ptp_operator/event/event.go
```diff
@@ -176,9 +204,9 @@ func (p *PTPEventState) UpdateCurrentEventState(c ClockState) ptp.SyncState {
 				"node": clockState.NodeName, "iface": alias}).Set(float64(v))
 		}
 		clockState.Metric = metrics
-		p.DependsOn[c.Process] = []*ClockState{clockState}
+		p.DependsOn[clockState.Process] = []*ClockState{clockState}
 	}
-
+	log.Infof("depends on end %s", p.PrintDependsOn())
```
jzding, line 209: remove temporary logs

plugins/ptp_operator/event/event.go
```diff
@@ -92,42 +95,54 @@ func (dt *DependingClockState) hasMetricHelp() map[string]string {
...
-			iface := ""
-			if d.IFace != nil {
-				iface = *d.IFace
-			}
-			if d.Metric == nil {
-				d.Metric = dep.hasMetric()
-				d.HelpText = dep.hasMetricHelp()
+			iface := "x"
+			if clockState.IFace != nil {
+				iface = *clockState.IFace
```
jzding, line 137: Shouldn't IFace be checked before entering the range c.Value for-loop? Otherwise if IFace is nil, then metrics with iface = 'x' will be created.

plugins/ptp_operator/event/event.go
```diff
@@ -240,24 +268,63 @@ func (p *PTPEventState) UnRegisterAllMetrics() {
...
-func (p *PTPEventState) DeleteAllMetrics() {
+func (p *PTPEventState) DeleteAllMetrics(m []*prometheus.GaugeVec) {
 	// write loop p.DependsOn
 	if p.DependsOn == nil {
 		return
 	}
 	for _, d := range p.DependsOn {
 		// write loop d.Metric
 		for _, dd := range d {
+			r := []rune(*dd.IFace)
```
jzding, line 279: check dd.IFace for nil?

plugins/ptp_operator/event/event.go
aneeshkp: Done

plugins/ptp_operator/event/event.go
aneeshkp: done

plugins/ptp_operator/event/event.go
aneeshkp: done

jzding (approved):
/lgtm

PR #297, opened 2024-03-27
fix cpu spike due to default in select

aneeshkp:
PR to fix file-watcher update introduced default for select that in turn introduced  cpu spike.
https://github.com/redhat-cne/cloud-event-proxy/pull/296
This PR fixes CPU spike by removing default from channel selector and introduce new channel to process config map after file delete operation is performed


jzding (approved):
/lgtm

PR #298, opened 2024-03-27
Enhance Consumer Application with Periodic Publisher Health Check

aneeshkp:
Previously, the consumer could only receive events if the producer was already running when the consumer started. This created a dependency issue, where a consumer restart was necessary if the producer was restarted or enabled after the consumer was already running.

This PR addresses this limitation by introducing a publisher health check. The consumer now performs a health check on the producer every second. This check ensures that the producer is healthy before subscribing to its events. This eliminates the need for consumer restarts due to producer restarts or enablement after consumer startup.

Benefits:
    Improved Consumer Resilience: This change enhances the consumer's robustness by making it independent of the producer's initial state.
    Simplified Management: No more consumer restarts are required solely due to producer lifecycle changes.
    Enhanced Reliability: By subscribing only to a healthy producer, the consumer avoids consuming potentially invalid or incomplete data.


jzding (approved):
/lgtm

PR #307, opened 2024-04-30
Fix consumer image and e2e-aws test

jzding:
- Fix event subscription in consumer example.
- Update consumer image and deployment.
- Remove amq from e2e test.


nishant-parekh (approved):
/lgtm

PR #321, opened 2024-07-05
OCPBUGS-36496: GM clock_state stuck in freerun

aneeshkp:
This is due to ts2phc reporting high offset. But clock state for ts2phc was showing as locked.
This PR  tries to update metrics accurately when ts2phc is high .


jzding (approved):
/lgtm

PR #324, opened 2024-07-17
Very high ts2phc offsets reported in linuxptp log while metrics show ts2phc clock_state is locked and GM clock_state is FREERUN

aneeshkp:
This is due to ts2phc reporting high offset. But clock state for ts2phc was showing as locked.
This PR tries to update metrics accurately when ts2phc is high .


jzding (approved):
/lgtm

PR #327, opened 2024-07-22
OCPBUGS-35587: T-GM - ts2phc process restart doesn't update PTP syncState when the process recovered

aneeshkp:
ts2phc process termination only triggers FREERUN and doesn't recover back .
This PR fixes this issue by reading process down status and handle ts2phc event

### DIFF
--- a/plugins/ptp_operator/metrics/metrics.go
+++ b/plugins/ptp_operator/metrics/metrics.go
@@ -181,6 +181,11 @@ func (p *PTPEventManager) ExtractMetrics(msg string) {
 			alias := getAlias(interfaceName)
 			// no event for nmeas status , change in GM will manage ptp events and sync states
 			UpdateNmeaStatusMetrics(processName, alias, status)
+		} else if strings.Contains(output, "process_status") &&
+			processName == ts2phcProcessName {
+			// do nothing processDown identifier will update  metrics and stats
+			// but prevent further from reading offsets
+			return
 		} else if strings.Contains(output, " offset ") { //  DPLL has Offset too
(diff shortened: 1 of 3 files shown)


plugins/ptp_operator/metrics/metrics.go
```diff
@@ -131,6 +131,7 @@ func (p *PTPEventManager) ExtractMetrics(msg string) {
 	//  check if ptp4l or phc2sys process are dead
 	// if process is down fire an event
 	if strings.Contains(output, ptpProcessStatusIdentifier) {
+		log.Infof("output %s", output)
```
jzding, line 134: is this temporary? too much logs.

plugins/ptp_operator/stats/stats.go
```diff
@@ -236,7 +236,7 @@ func (s *Stats) GetStateState(processName string, iface *string) (ptp.SyncState,
 				return d[0].State, nil
 			}
 			for _, state := range d {
-				if *state.IFace == *iface {
+				if iface != nil && *state.IFace == *iface {
```
jzding, line 239: should check state != nil

aneeshkp: No required , ranging inputState (state) struct and iface is a pointer .

plugins/ptp_operator/metrics/metrics.go
aneeshkp: yes , it was debug log . Thanks will remove it

jzding (approved):
/lgtm

PR #338, opened 2024-08-14
OCPBUGS-35799: OCPBUGS-38284: clean up subscriptions when lost connection to consumer

jzding:
Fixe both v1 and v2 cleanup: reset fail count when a event is successfully delivered.

Also added API support for deleting all subscriptions:
    DELETE /api/ocloudNotifications/v2/subscriptions

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -18,8 +18,10 @@ package main
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"flag"
 	"fmt"
+	"net"
 	"net/http"
 	"os"
 	"os/signal"
@@ -233,7 +235,7 @@ func ProcessOutChannel(wg *sync.WaitGroup, scConfig *common.SCConfiguration) {
 			if pub.EndPointURI != nil {
 				log.Debugf("posting acknowledgment with status: %s to publisher: %s", status, pub.EndPointURI)
 				restClient := restclient.New()
-				_ = restClient.Post(pub.EndPointURI,
+				restClient.Post(pub.EndPointURI,
 					[]byte(fmt.Sprintf(`{eventId:"%s",status:"%s"}`, pub.ID, status)))
 			}
 		}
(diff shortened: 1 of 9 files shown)


cmd/main.go
```diff
@@ -286,16 +289,45 @@ func ProcessOutChannel(wg *sync.WaitGroup, scConfig *common.SCConfiguration) {
...
+										// Capture DNS error "lookup consumer-events-subscription-service.cloud-events.svc.cluster.local: no such host"
+										// or timeout error "context deadline exceeded (Client.Timeout exceeded while awaiting headers)"
+										if errors.As(err, &dnsError) || os.IsTimeout(err) {
+											// has subscriber failed to connect for 10 times delete the subscribers
+											if scConfig.SubscriberAPI.IncFailCountToFail(clientID) {
+												log.Errorf("connection lost for address %s, proceed to clean up subscription", d.Address)
+												if numSubDeleted, err = scConfig.SubscriberAPI.DeleteAllSubscriptionsForClient(clientID); err != nil {
+													log.Errorf("failed to delete all subscriptions for client %s: %s", clientID.String(), err.Error())
+												} else {
+													cleanupConfigMap(d.ClientID)
```
aneeshkp, line 309: does this clean up inmemory database of the publisher from the store map ?

cmd/main.go
```diff
@@ -286,16 +289,45 @@ func ProcessOutChannel(wg *sync.WaitGroup, scConfig *common.SCConfiguration) {
...
+										log.Errorf("error posting event at %s : %s", endPointURI, err)
+										localmetrics.UpdateEventReceivedCount(d.Address, localmetrics.FAILED)
+
+										// POST return DNSError if server is not reachable
+										var dnsError *net.DNSError
+										// Capture DNS error "lookup consumer-events-subscription-service.cloud-events.svc.cluster.local: no such host"
+										// or timeout error "context deadline exceeded (Client.Timeout exceeded while awaiting headers)"
+										if errors.As(err, &dnsError) || os.IsTimeout(err) {
+											// has subscriber failed to connect for 10 times delete the subscribers
+											if scConfig.SubscriberAPI.IncFailCountToFail(clientID) {
```
aneeshkp, line 304: do you dues the threshold to wait few tries before cleaning up ?
log.Errorf("client %s not responding, waiting %d times before marking to delete subscriber",
						clientAddress, h.subscriberAPI.FailCountThreshold()-h.subscriberAPI.GetFailCount(clientID))

cmd/main.go
jzding: added cleanup

jzding: in-memory db is only cleaned up when deleting all subs for all clients. In this case for client timeout, only deleting from SubscribnerAPI will be enough. The events will be stopped for this client.

aneeshkp:
@jzding failing linter and unit test

aneeshkp (approved):
/lgtm

PR #353, opened 2024-09-11
OCPBUGS-39218: fix double / in ResourceAddress

jzding:


plugins/ptp_operator/metrics/manager.go
```diff
@@ -260,7 +260,10 @@ func (p *PTPEventManager) GetPTPEventsData(state ptp.SyncState, ptpOffset int64,
 		return nil
 	}
 	// /cluster/xyz/ptp/CLOCK_REALTIME this is not address the event is published to
-	eventSource := fmt.Sprintf(p.resourcePrefix, p.nodeName, fmt.Sprintf("/%s", source))
+	if !strings.HasPrefix(source, "/") {
```
aneeshkp, line 263: use path.join 
eventSource := fmt.Sprintf("%s", path.Join(p.resourcePrefix, p.nodeName, source))

jzding: can't simply use path.Join, this is what came out. 
`/cluster/node/%s%s/cnfdg4.ptp.eng.rdu2.dc.redhat.com/ptp-not-set`
added more changes

aneeshkp (approved):
/lgtm

PR #358, opened 2024-09-23
OCPBUGS-19838:PTP HW Clock Class after disabling NMEA sentences - Handle T-GM holdover

aneeshkp:
This PR  handles T-GN holdover fixes introduced in https://github.com/openshift/linuxptp-daemon/pull/367
New metrics getting generated for ts2phc  in holdover state


jzding (approved):
/lgtm

PR #363, opened 2024-10-15
Remove hard coded replacement of api addr in consumer deployment 

aneeshkp:
This PR removes the hardcoded replacement of apiAddr in the consumer application when switching from V1 to V2. Instead of relying on manual hardcoding, the application will now dynamically use the first publisher from the provided list via the --http-event-publishers argument.

The new implementation will fetch the first publisher from the following format :
 --http-event-publishers=ptp-event-publisher-service-NODE_NAME.openshift-ptp.svc.cluster.local:9043


jzding (approved):
/lgtm

PR #368, opened 2024-10-31
add log-events.py script

jzding:
Usaeg: 
`oc -n cloud-events logs -f deployment/cloud-consumer-deployment -c cloud-event-consumer | python3 hack/log-events.py`

Results:
⬅️ 2024-10-31T14:23:01.6652Z: event.sync.sync-status.os-clock-sync-state-change
⬅️ 2024-10-31T14:24:01.6618Z: event.sync.ptp-status.ptp-state-change


aneeshkp (approved):

hack/log-events.py
```diff
@@ -0,0 +1,56 @@
...
+    PUSH = "PUSH"
+    PULL = "PULL"
+
+def process_log(log_message, push_pull):
+    try:
+        # Remove escape characters
+        log_message = log_message.replace('\\"', '"')
+
+        # Extract JSON part from the log message
+        json_part = re.search(r'{.*}', log_message).group()
```
aneeshkp, line 17: you should check if this  return result

aneeshkp: json_match = re.search(r'{.*}', log_message)
        if not json_match:
            print("No JSON part found in log message.")
            return

aneeshkp (changes requested):
please see comments

hack/log-events.py
jzding: updated

aneeshkp (approved):
/lgtm

PR #369, opened 2024-11-04
OCPBUGS-44208: Fix Holdover Event Triggering for T-GM and Minimize Redundant Clock Class Events:

aneeshkp:
This PR addresses two issues:

1.  Fix Holdover Event Triggering for T-GM :
        Issue: When the T-GM (Telecom Grandmaster) entered holdover mode, the associated events were not generated, although metrics were successfully created.
        Resolution: This PR ensures that event generation is correctly triggered for the holdover state, aligning events with the metrics generated.

2. Eliminate Duplicate Clock Class Events:
        Issue: Multiple events were being generated for the same clock class value, causing redundancy.
        Resolution: Duplicate events for the same clock class value have been removed, streamlining event reporting and reducing noise.


jzding:
could you add unit test for the event?

jzding (approved):
/lgtm

PR #373, opened 2024-11-26
[release-4.16] OCPBUGS-44900: PTP operator Event Notification API breaking change within 4.16 release

aneeshkp:
Fix Backward Compatibility for PTP Clock-Class Event Subscription

The problem arises because the V1 version of PTP events subscribed to the resource address /sync/ptp-status/ptp-clock-class-change, which was updated to /sync/ptp-status/clock-class in V2 as per O-RAN spec.
As a result, older versions referencing /sync/ptp-status/ptp-clock-class-change failed to subscribe to clock-class events.

This PR resolves the issue by introducing a new enum type, PtpClockClassV1, which refers to the older event resource.
The solution ensures backward compatibility for V1 while adhering to the O-RAN 7.2.3.10 specification for the PtpClockClass notification.

Key Changes:
In Cloud-event-proxy , manages different clock type subscription names by detecting API version
```
if v1Api {
                  subscribeTo[string(ptpEvent.PtpClockClassChange)] = string(ptpEvent.PtpClockClassV1)
          }

```
In redhat-cne/sd-go:
sdk-go updated by adding PtpClockClassV1 to support the older /sync/ptp-status/ptp-clock-class-change resource for V1 compatibility.
Retained the PtpClockClass resource for the updated /sync/ptp-status/clock-class to align with the O-RAN specification.
Updated Definitions:
// O-RAN 7.2.3.10
// PtpClockClass notification is generated when the clock-class changes.
PtpClockClass EventResource = "/sync/ptp-status/clock-class"

// Support V1
// PtpClockClassV1 notification is generated when the clock-class changes for V1.
PtpClockClassV1 EventResource = "/sync/ptp-status/ptp-clock-class-change"

This change ensures that both V1 and the current versions of PTP events can correctly subscribe to clock-class notifications without breaking functionality for older versions.

### DIFF
--- a/examples/consumer/main.go
+++ b/examples/consumer/main.go
@@ -107,7 +107,7 @@ func main() {
 		log.Infof("apiVersion=%s, updated apiAddr=%s, apiPath=%s", apiVersion, apiAddr, apiPath)
 	}
 
-	subscribeTo := initSubscribers(consumerType)
+	subscribeTo := initSubscribers(consumerType, isV1Api)
 	var wg sync.WaitGroup
 	wg.Add(1)
 	go server() // spin local api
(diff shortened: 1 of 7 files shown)


jzding (approved):
/lgtm

PR #380, opened 2024-12-18
OCPBUGS-45866 : Fix GM State Transition Event Generation

vitus133:
This is a cherrypick from #379
Previously, when the Grandmaster (GM) transitioned to the HOLDOVER state, an event was generated with the HOLDOVER status. However, subsequent state transitions from HOLDOVER to FREERUN or LOCKED did not trigger any events. This bug fix ensures that events are now correctly generated for transitions from HOLDOVER to FREERUN or LOCKED states.


jzding:
the change in [plugins/ptp_operator/metrics/metrics_test.go](https://github.com/redhat-cne/cloud-event-proxy/pull/379/files#diff-60724f4e87547bf095e9209fe896077e8d07da47f038c0083c6772b040135df8) is missing. It will be great to add unit test coverage.

jzding (approved):
/lgtm

PR #414, opened 2025-02-11
Obsolete - Drop support for APIv1

rdavid:
- Exit on attempt to run APIv1.
- Remove deprecated code.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -86,7 +86,7 @@ func main() {
 	flag.StringVar(&storePath, "store-path", ".", "The path to store publisher and subscription info.")
 	flag.StringVar(&transportHost, "transport-host", "http://ptp-event-publisher-service-NODE_NAME.openshift-ptp.svc.cluster.local:9043", "The transport bus hostname or service name.")
 	flag.IntVar(&apiPort, "api-port", 8089, "The address the rest api endpoint binds to.")
-	flag.StringVar(&apiVersion, "api-version", "1.0", "The address the rest api endpoint binds to.")
+	flag.StringVar(&apiVersion, "api-version", "2.0", "The address the rest api endpoint binds to.")
 	flag.StringVar(&httpEventPublisher, "http-event-publishers", "", "Comma separated address of the publishers available.")
 
 	flag.Parse()
@@ -159,6 +159,10 @@ func main() {
 
 	pluginHandler = plugins.Handler{Path: "./plugins"}
 	isV1Api = common.IsV1Api(scConfig.APIVersion)
+	if isV1Api {
+		log.Fatalf("APIv1 is not supported since OCP v4.19: %s.", scConfig.APIVersion)
+	}
+
 	// make PubSub REST API accessible by event service ptp-event-publisher-service-NODE_NAME
 	if !isV1Api {
 		// switch between Internal API port and PubSub port
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -71,7 +71,6 @@ var (
 	pluginHandler           plugins.Handler
 	nodeName                string
 	namespace               string
-	isV1Api                 bool
 )
 
 func main() {
@@ -150,16 +149,14 @@ func main() {
 	}()
 
 	pluginHandler = plugins.Handler{Path: "./plugins"}
-	isV1Api = common.IsV1Api(scConfig.APIVersion)
-	if isV1Api {
+	if common.IsV1Api(scConfig.APIVersion) {
 		log.Fatal("APIv1 is not supported since OCP v4.19.")
 	}
 	log.Infof(
-		"REST API config: version=%s, port=%d, path=%s, TransportHost.Port=%d.",
+		"REST API config: version=%s, port=%d, path=%s.",
 		scConfig.APIVersion,
 		scConfig.APIPort,
-		scConfig.APIPath,
-		scConfig.TransportHost.Port)
+		scConfig.APIPath)
 
 	// Enable pub/sub services
 	err = common.StartPubSubService(scConfig)
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -67,10 +67,7 @@ var (
 	statusChannelBufferSize = 50
 	scConfig                *common.SCConfiguration
 	metricsAddr             string
-	apiPath                 = "/api/ocloudNotifications/v1/"
-	// Event API Version 2 is O-RAN V3.0 Compliant
-	apiPathV2          = "/api/ocloudNotifications/v2/"
-	httpEventPublisher string
+	apiPath                 = "/api/ocloudNotifications/v2/"
 	pluginHandler      plugins.Handler
 	nodeName           string
 	namespace          string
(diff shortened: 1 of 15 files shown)


cmd/main.go
```diff
@@ -86,7 +84,7 @@ func main() {
 	flag.StringVar(&storePath, "store-path", ".", "The path to store publisher and subscription info.")
 	flag.StringVar(&transportHost, "transport-host", "http://ptp-event-publisher-service-NODE_NAME.openshift-ptp.svc.cluster.local:9043", "The transport bus hostname or service name.")
 	flag.IntVar(&apiPort, "api-port", 8089, "The address the rest api endpoint binds to.")
-	flag.StringVar(&apiVersion, "api-version", "1.0", "The address the rest api endpoint binds to.")
+	flag.StringVar(&apiVersion, "api-version", "2.0", "The address the rest api endpoint binds to.")
 	flag.StringVar(&httpEventPublisher, "http-event-publishers", "", "Comma separated address of the publishers available.")
```
jzding, line 88: httpEventPublisher is no longer needed. It was used for the sidecar on consumer side, but with v2 there is no longer sidecar on consumer side.

cmd/main.go
```diff
@@ -86,7 +84,7 @@ func main() {
 	flag.StringVar(&storePath, "store-path", ".", "The path to store publisher and subscription info.")
 	flag.StringVar(&transportHost, "transport-host", "http://ptp-event-publisher-service-NODE_NAME.openshift-ptp.svc.cluster.local:9043", "The transport bus hostname or service name.")
```
jzding, line 85: transportHost should be renamed apiHost

cmd/main.go
```diff
@@ -86,7 +84,7 @@ func main() {
...
 	flag.IntVar(&apiPort, "api-port", 8089, "The address the rest api endpoint binds to.")
```
jzding, line 86: `8089` is the v1 internal API port that we no longer need. We can set apiPort to `9043` which is the v2 API port.

cmd/main.go
jzding: On second thought, maybe it is best to keep the naming of these flags for now, since changing them would require migration path and lots of documentation updates.
I think for now we can use transport-host to get both apiHost and apiPort (9043), keep the api-port flag as is but not actually use its value.

jzding: We can change the name later in different PRs since it would require changes in ptp-operator and test scripts.

hack/common/consumer.sh
```diff
@@ -72,7 +72,7 @@ spec:
             - "--store-path=/store"
             - "--transport-host=$transport_host"
             - "--http-event-publishers=$http_event_publishers"
-            - "--api-port=9095"
+            - "--api-port=9043"
```
jzding, line 75: for v2 there should not be a cloud-event-proxy sidecar. Please refer to https://github.com/redhat-cne/cloud-event-proxy/tree/main/examples/manifests/v2

test/cne/cne.go
```diff
@@ -92,21 +92,6 @@ var _ = ginkgo.Describe("validation", func() {
...
-				ginkgo.By("Checking event api is healthy")
-				gomega.Eventually(func() string {
-					buf, _ := pods.ExecCommand(testclient.Client, producerPod, testutils.EventProxyContainerName, []string{"curl", "127.0.0.1:9095/api/ocloudNotifications/v1/health"})
-					return buf.String()
-				}, 5*time.Minute, 5*time.Second).Should(gomega.ContainSubstring("OK"),
-					"Event API is not in healthy state")
-
-				ginkgo.By("Checking mock publisher is created")
-				gomega.Eventually(func() string {
-					buf, _ := pods.ExecCommand(testclient.Client, producerPod, testutils.EventProxyContainerName, []string{"curl", "127.0.0.1:9095/api/ocloudNotifications/v1/publishers"})
```
jzding, line 105: should replace these with v2 API equivalent

test/cne/cne.go
```diff
@@ -136,21 +121,6 @@ var _ = ginkgo.Describe("validation", func() {
...
-				ginkgo.By("Checking event api is healthy")
-				gomega.Eventually(func() string {
-					buf, _ := pods.ExecCommand(testclient.Client, consumerPod, testutils.EventProxyContainerName, []string{"curl", "127.0.0.1:9095/api/ocloudNotifications/v1/health"})
-					return buf.String()
-				}, 5*time.Minute, 5*time.Second).Should(gomega.ContainSubstring("OK"),
-					"Event API is not in healthy state")
-
-				ginkgo.By("Checking mock subscription is created")
-				gomega.Eventually(func() string {
-					buf, _ := pods.ExecCommand(testclient.Client, consumerPod, testutils.EventProxyContainerName, []string{"curl", "127.0.0.1:9095/api/ocloudNotifications/v1/subscriptions"})
```
jzding, line 149: should replace these with v2 API equivalent

hack/common/consumer.sh
```diff
@@ -53,8 +53,8 @@ spec:
           imagePullPolicy: Always
           args:
             - "--local-api-addr=127.0.0.1:9089"
```
jzding, line 55: this should change to consumer-events-subscription-service.${NamespaceConsumerTesting}.svc.cluster.local:9043

jzding: since this is the endpoint URI that producer API need to reach and it is in a different pod than the producer, so can not use local host.

rdavid: Who will set `NamespaceConsumerTesting`, and where will it be set?

hack/common/consumer.sh
rdavid: I don't understand. The sidecar has been removed.

hack/common/consumer.sh
jzding: it is defined here https://github.com/redhat-cne/cloud-event-proxy/blob/main/hack/common/common.sh#L6

jzding: Namespace `cloud-events` is for consumer example manifests. In the e2e test it uses a different namespace.

cmd/main.go
```diff
@@ -158,39 +149,21 @@ func main() {
...
-		transportEnabled = false
-	}
-
-	// if all transport types failed then process internally
-	if !transportEnabled && isV1Api {
-		log.Errorf("No transport is enabled for sending events %s", scConfig.TransportHost.String())
-		wg.Add(1)
-		go ProcessInChannel(&wg, scConfig)
+	if common.IsV1Api(scConfig.APIVersion) {
+		log.Fatal("APIv1 is not supported since OCP v4.19.")
```
jzding, line 153: "REST API v1 is no longer supported. Please update api-version to 2.0"

hack/common/cne.sh
```diff
@@ -51,8 +51,8 @@ spec:
             - "--metrics-addr=127.0.0.1:9091"
             - "--store-path=/store"
             - "--transport-host=$transport_host"
-            - "--api-version=2.0"
             - "--api-port=9095"
```
jzding, line 54: In main.go you are using api-port instead of port retrieved from transport-host, so this need to be changed to 9043

PR #441, opened 2025-03-21
CNF-16979: Remove Event API v1 support from consumer app

jzding:
- Remove Event API v1 from consumer app
- Remove v1 manifests from consumer
- With this change the `--api-addr` and `--api-version` flags for consumer are obsolete.


examples/manifests/kustomization.yaml
```diff
@@ -3,16 +3,12 @@ kind: Kustomization
...
   name: cloud-consumer-deployment
 images:
 - name: cloud-event-consumer
-  newName: quay.io/redhat-cne/cloud-event-consumer
-  newTag: latest
-- name: cloud-event-sidecar
-  newName: quay.io/openshift/origin-cloud-event-proxy
-  newTag: latest
+  newName: quay.io/jacding/cloud-event-consumer
+  newTag: remove-v1
```
rdavid, line 14: Is it a leftover from private testing?

rdavid (approved):

examples/manifests/kustomization.yaml
jzding: yes, thanks. fixing this in a new PR https://github.com/redhat-cne/cloud-event-proxy/pull/442/files

PR #453, opened 2025-03-27
enable multi-arch consumer image

jzding:
Replace quay.io Build Trigger with Github workflow.
Trigger the build and push of consumer image when a PR is merged to `main` or `release-4.18` branches.

For PR merged to main branch, this workflow will create the following images:

- quay.io/redhat-cne/cloud-event-consumer:latest (multi-arch)
- quay.io/redhat-cne/cloud-event-consumer:latest-arm64
- quay.io/redhat-cne/cloud-event-consumer:latest-amd64

For PR merged to release-4.18 branch, this workflow will create the following images:

- quay.io/redhat-cne/cloud-event-consumer:release-4.18 (multi-arch)
- quay.io/redhat-cne/cloud-event-consumer:release-4.18-arm64
- quay.io/redhat-cne/cloud-event-consumer:release-4.18-amd64

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,0 +1,76 @@
+name: Build and Push Consumer Image
+on:
+  push:
+    branches: [ main, release-4.18 ]
+env:
+  VERSION: ${{ github.ref_name == 'main' && 'latest' || github.ref_name == 'release-4.18' && 'release-4.18' }}
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+          - platform: linux/amd64
+            arch: amd64
+          - platform: linux/arm64
+            arch: arm64
+    steps:
+      - name: Debug
+        run: |
+          echo "VERSION: ${{ env.VERSION }}"
+          echo "GITHUB_REF_NAME: ${{ github.ref_name }}"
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yaml
```diff
@@ -0,0 +1,76 @@
+name: Build and Push Consumer Image
+on:
+  push:
+    branches: [ main, release-4.18 ]
+env:
+  VERSION: ${{ github.ref_name == 'main' && 'latest' || github.ref_name == 'release-4.18' && 'release-4.18' }}
```
edcdavid, line 6: what about just returning the branch name, this way it does not need to be updated later:
```
 VERSION: ${{ github.ref_name == 'main' && 'latest' || github.ref_name }}
```

.github/workflows/build.yaml
```diff
@@ -0,0 +1,76 @@
+name: Build and Push Consumer Image
+on:
+  push:
+    branches: [ main, release-4.18 ]
```
edcdavid, line 4: Is it ok to run on all branches? This way we don't need to update this later:
```
branches:
  - "*"
```

edcdavid (approved):

.github/workflows/build.yaml
jzding: we do not want to run on all branches. 4.18 is the last version to support v1 events. For v2 events we always use main branch.

PR #459, opened 2025-04-03
OCPBUGS-54543: Data race in Server.status access causes instability in cloud-event-proxy

aneeshkp:
Update go.mod to use rest-api v1.23.1, which includes a fix for a race condition when reading the server status.

api fix: [redhat-cne/rest-api#90](https://github.com/redhat-cne/rest-api/pull/90)


jzding (approved):
/lgtm

PR #469, opened 2025-04-14
OCPBUGS-45680: add error handling in configmap and consumer app subscription

jzding:
consumer changes:
- re-subscribe in case of non-recoverable error at server side, for example lost of persist data in configmap
- add git commit ID in consumer image

sidecar changes:
- add retry in configMap creation
- add trace and error handling to configMap creation and updating

### DIFF
--- a/examples/consumer/main.go
+++ b/examples/consumer/main.go
@@ -72,9 +72,12 @@ var (
 	// map to track if subscriptions were created successfully for each publisher service
 	subscribed = make(map[string]bool)
 	subs       []*pubsub.PubSub
+	// Git commit of current build set at build time
+	GitCommit = "Undefined"
 )
 
 func main() {
+	fmt.Printf("Git commit: %s\n", GitCommit)
 	common.InitLogger()
 	flag.StringVar(&localAPIAddr, "local-api-addr", "localhost:8989", "The address the local api binds to .")
 	flag.StringVar(&apiPath, "api-path", "/api/ocloudNotifications/v2/", "The rest api path.")
--- a/examples/consumer/main.go
+++ b/examples/consumer/main.go
@@ -173,6 +173,34 @@ func deleteAllSubscriptions() {
 	}
 }
 
+// listSubscriptions lists all subscriptions
+// and returns true if there are any subscriptions
+func listSubscriptions() bool {
+	url := &types.URI{URL: url.URL{Scheme: "http",
+		Host: apiAddr,
+		Path: fmt.Sprintf("%s%s", apiPath, "subscriptions")}}
+	rc := restclient.New()
+
+	var subs = []pubsub.PubSub{}
+	var subB []byte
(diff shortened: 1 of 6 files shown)


pkg/restclient/client.go
```diff
@@ -168,24 +168,24 @@ func (r *Rest) Delete(url *types.URI) int {
 }
 
 // Get  http request
-func (r *Rest) Get(url *types.URI) (int, string) {
+func (r *Rest) Get(url *types.URI) (int, []byte, error) {
 	ctx := context.Background()
 	ctx, cancel := context.WithCancel(ctx)
```
vitus133, line 173: nit
```suggestion
	ctx, cancel := context.WithCancel(context.Background())
```

pkg/restclient/client.go
```diff
@@ -168,24 +168,24 @@ func (r *Rest) Delete(url *types.URI) int {
...
 	defer cancel()
 	request, err := http.NewRequestWithContext(ctx, "GET", url.String(), nil)
 	if err != nil {
 		log.Errorf("error creating post request %v", err)
-		return http.StatusBadRequest, fmt.Sprintf("error creating post request %v", err)
+		return http.StatusBadRequest, nil, fmt.Errorf("error creating post request %v", err)
```
vitus133, line 178: All the error messages here talk about "post request", but actually this is a "get" request. The above function is "delete", but the comments are "post" again

examples/consumer/main.go
```diff
@@ -178,7 +207,14 @@ RETRY:
 			if !healthOK {
 				deleteAllSubscriptions()
 				log.Info("delete all subscriptions due to publisher health failure")
-				return
+				goto RETRY
```
vitus133, line 210: When we retry, is there a theoretical possibility of multiple retries? If yes, could it be beneficial to introduce a sleep before retrying?

examples/consumer/main.go
```diff
@@ -72,9 +72,12 @@ var (
...
 	subscribed = make(map[string]bool)
 	subs       []*pubsub.PubSub
+	// Git commit of current build set at build time
+	GitCommit = "Undefined"
 )
 
 func main() {
+	fmt.Printf("Git commit: %s\n", GitCommit)
 	common.InitLogger()
 	flag.StringVar(&localAPIAddr, "local-api-addr", "localhost:8989", "The address the local api binds to .")
```
vitus133, line 82: Just a thought - could it be beneficial to use a flag or an environment variable instead of the hardcoded port?

vitus133 (commented):
Left a couple of comments, otherwise looks good!

examples/consumer/main.go
jzding: deleteAllSubscriptions() will set the `subOK` to false, so retry will go to different path. If the new path failed it will retry after 30 seconds.

vitus133 (approved):

PR #470, opened 2025-04-15
OCPBUGS-54984: Update Sync State Logic for Downstream Slave When All HA Interfaces Are Down

aneeshkp:
In OC  configuration with single port or dual follower port , taking down both interfaces leads to unexpected behavior:The interface role transitions from SLAVE to LISTENING, but the system does not correctly reflect this loss of sync.

 When any port transitions from SLAVE to FAULTY, it will go through the HOLDOVER state. This should resolve the issue, so no special handling is needed — all transitions will consistently pass through HOLDOVER.
For example:
If port 1 becomes FAULTY and port 2 transitions to SLAVE, HOLDOVER should be released.
If both ports transition to FAULTY, the system will move to FREERUN via the HOLDOVER state.


plugins/ptp_operator/metrics/logparser.go
```diff
@@ -297,7 +297,7 @@ func extractPTP4lEventState(output string, isFollowerOnly bool) (portID int, rol
 		strings.Contains(output, "SLAVE to UNCALIBRATED") ||
 		strings.Contains(output, "MASTER to UNCALIBRATED on RS_SLAVE") ||
 		strings.Contains(output, "LISTENING to UNCALIBRATED on RS_SLAVE") ||
-		(strings.Contains(output, "FAULTY to LISTENING on INIT_COMPLETE") && !isFollowerOnly) { // added to manage two port case so its not breaking
+		strings.Contains(output, "FAULTY to LISTENING on INIT_COMPLETE") { // added to manage two port case so its not breaking
```
edcdavid, line 300: Shouldn't this transition lead to LISTENING? Why is it FAULTY? the same transition is on line 310. Imho, we should process the LISTENING condition to also trigger the HOLDOVER if needed

edcdavid: the role is exposed via metrics, so in this case a LISTENING role would be FAULTY. Would be simpler to not change the roles. The LISTENING is a fault only if there is not other port that is a SLAVE, so in the general case LISTENING is not necessarily FAULTY

edcdavid: We can work on this refactor in a new PR. I will accept these changes.

edcdavid (approved):

PR #483, opened 2025-04-17
[release-4.15] fix ut and lint

jzding:


examples/producer/main.go
```diff
@@ -99,7 +99,11 @@ RETRY:
 					},
 					},
 				}
-				data.SetVersion("v1") //nolint:errcheck
+
+				if err := data.SetVersion("v1"); err != nil {
```
aneeshkp, line 103: we should ignore this  error , so that events are delivered
_=data.SetVersion("v1")

jzding: if the error did occur and we ignore it, the event would be failed to sent later due to marshalling error. 
This is test code.

aneeshkp: ok, this is sample .

aneeshkp (approved):
/lgtm

PR #488, opened 2025-04-23
OCPBUGS-55391: reduce event logs

jzding:
Reduce logs in consumer sidecar from 3 to 1 for each v1 event pull. See detail in https://issues.redhat.com/browse/OCPBUGS-55264


aneeshkp (approved):
/approve

PR #496, opened 2025-05-06
Reduce clock class event noise by checking previous clock class state 

aneeshkp:
Optimize ClockClass Event Publishing and Filter Non-ptp4l Port Events
This PR reduces clock class frequency that was introduced in this PR https://github.com/openshift/linuxptp-daemon/pull/427  Under linuxptp daemon repo 

This PR introduces two key improvements to the PTP metrics and event publishing logic:

    Avoid Redundant ClockClass Updates and Events
    The SetClockClass() and PublishClockClassEvent() calls are now gated by a check to ensure that the ClockClass value has actually changed. This prevents unnecessary updates and repeated event publishing when the clock class remains the same.

    Filter Out Non-ptp4l Port Events
    Modified the port event handling logic to process "port" logs only from the ptp4l process, ignoring outputs from other processes like phc2sys. This reduces noise and ensures more accurate PTP state tracking.


josephdrichard (approved):
/lgtm

PR #505, opened 2025-05-13
[release-4.19] OCPBUGS-56113: Extract listening state when transitioning from faulty

edcdavid:
FAULTY to LISTENING transition is reported as a FAULTY state instead of LISTENING


aneeshkp (approved):
/lgtm

PR #523, opened 2025-05-22
Stop using IsV1Api

rdavid:


rdavid:
@jzding The comment has been addressed.

PR #535, opened 2025-06-11
OCPBUGS-57063: fix concurrent map read in getCurrentStatOverrideFn

jzding:
- Use concurrent map read eventManager.GetStats() in getCurrentStatOverrideFn
- fix error log when event data not found


aneeshkp (approved):
lgtm

PR #567, opened 2025-08-01
OCPBUGS-59969: Handle missing event data in getCurrentState with 404 response

jzding:
Previously, getCurrentState returned a FREERUN state with placeholder ResourceAddresses ("event-not-found" or "ptp-not-set") when event data was unavailable, for example when the event socket wasn't ready, risking cell site outages. This update ensures a 404 Not Found status is returned instead, accurately signaling the absence of valid event data.

Add validation to prevent Stats.ClockClass() from returning an uninitialized value (0), which could result in unintended events with clockClass=0.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,7 +11,7 @@ require (
 	github.com/onsi/ginkgo v1.16.5
 	github.com/onsi/gomega v1.23.0
 	github.com/prometheus/client_golang v1.14.0
-	github.com/redhat-cne/rest-api v1.23.3
+	github.com/redhat-cne/rest-api v1.23.5
 	github.com/redhat-cne/sdk-go v1.23.4
 	github.com/sirupsen/logrus v1.9.0
 	github.com/stretchr/testify v1.8.1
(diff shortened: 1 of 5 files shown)


plugins/ptp_operator/stats/stats.go
```diff
@@ -178,11 +177,13 @@ func (s *Stats) SetDelay(val int64) {
 
 // SetLastOffset ... set last offset value
 func (s *Stats) SetLastOffset(val int64) {
+	s.lastOffsetSet = true
 	s.lastOffset = val
 }
 
 // SetClockClass ... set last clock class value
 func (s *Stats) SetClockClass(val int64) {
+	s.clockClassSet = true
```
aneeshkp, line 186: why can't use clock class ==0 as not set

plugins/ptp_operator/metrics/ptp4lParse.go
```diff
@@ -47,7 +47,8 @@ func (p *PTPEventManager) ParsePTP4l(processName, configName, profileName, outpu
 
 			ClockClassMetrics.With(prometheus.Labels{
 				"process": processName, "node": ptpNodeName}).Set(clockClass)
-			if ptpStats[master].ClockClass() != int64(clockClass) {
+			// if clock class has never been set, or it is not the same as the one reported by ptp4l, update the clock class
+			if ptpStats[master].ClockClass() != int64(clockClass) || !ptpStats[master].IsClockClassSet() {
```
aneeshkp, line 51: or clockass ==0 which is considered as undefined

PR #571, opened 2025-08-06
[release-4.18] OCPBUGS-59984: Handle missing event data in getCurrentState with 404/400 response

jzding:
Previously, getCurrentState returned a FREERUN state with placeholder ResourceAddresses ("event-not-found" or "ptp-not-set") when event data was unavailable, for example when the event socket wasn't ready, risking cell site outages. This update ensures a 404(v2)/400(v1) status is returned instead, accurately signaling the absence of valid event data.

Add validation to prevent Stats.ClockClass() from returning an uninitialized value (0), which could result in unintended events with clockClass=0.


aneeshkp (approved):
/lgtm

PR #572, opened 2025-08-06
OCPBUGS-60191: Fix PTP interface role assignment and metrics generation on container restart[release-4.18]

aneeshkp:
Assisted by : Cursor AI
Problem
When the PTP container restarts, it fails to properly pick up config information and complete metrics for interface roles. This results in:
Interface roles remain UNKNOWN: The validLogToProcess function fails validation because interfaces have incomplete role assignments
Missing interface_role metrics: Metrics are not generated for interface roles during config processing
Log processing skipped: The system skips processing PTP logs due to validation failures, preventing proper metrics collection. 

**KNOWN ISSUE:**  ```
The interface role will be based on ptp config , if anything changed , side car  won't know. 
Future fix to update interface role from the daemon will fix this problem 
```

### DIFF
--- a/plugins/ptp_operator/metrics/manager.go
+++ b/plugins/ptp_operator/metrics/manager.go
@@ -292,17 +292,13 @@ func (p *PTPEventManager) publishSyncEEvent(syncState ptp.SyncState, source stri
 
 // GetPTPEventsData ... get PTP event data object
 func (p *PTPEventManager) GetPTPEventsData(state ptp.SyncState, ptpOffset int64, source string, eventType ptp.EventType) *ceevent.Data {
-	// create an event
-	if state == "" {
-		return nil
-	}
 	// /cluster/xyz/ptp/CLOCK_REALTIME this is not address the event is published to
 	eventSource := path.Join(p.resourcePrefix, p.nodeName, source)
 	data := ceevent.Data{
 		Version: ceevent.APISchemaVersion,
 		Values:  []ceevent.DataValue{},
 	}
-	if eventType != ptp.PtpClockClassChange && eventType != ptp.SynceClockQualityChange {
+	if eventType != ptp.PtpClockClassChange && eventType != ptp.SynceClockQualityChange && state != "" {
 		data.Values = append(data.Values, ceevent.DataValue{
 			Resource:  eventSource,
 			DataType:  ceevent.NOTIFICATION,
--- a/plugins/ptp_operator/ptp_operator_plugin.go
+++ b/plugins/ptp_operator/ptp_operator_plugin.go
@@ -424,16 +424,56 @@ func processPtp4lConfigFileUpdates() {
 				var ptpInterfaces []*ptp4lconf.PTPInterface
 				for index, ptpInterface := range newInterfaces {
 					role := ptpTypes.UNKNOWN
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: plugins/ptp_operator/ptp_operator_plugin.go [hunk 1]
```diff
@@ -607,7 +628,7 @@ func processMessages(c net.Conn) {
 	}
 }
 
-// HasEqualInterface checks if configmap  changes has same interface
+// HasEqualInterface checks if configmap changes has same interface
```
josephdrichard, line 631: If you're changing the comment, can you make it more clear?  I'm not really sure what it's doing from this comment.

plugins/ptp_operator/metrics/manager.go
```diff
@@ -293,7 +293,7 @@ func (p *PTPEventManager) publishSyncEEvent(syncState ptp.SyncState, source stri
 // GetPTPEventsData ... get PTP event data object
 func (p *PTPEventManager) GetPTPEventsData(state ptp.SyncState, ptpOffset int64, source string, eventType ptp.EventType) *ceevent.Data {
 	// create an event
-	if state == "" {
+	if state == "" && eventType != ptp.PtpClockClassChange && eventType != ptp.SynceClockQualityChange {
```
josephdrichard, line 296: Do we need to do this check 3 times?

plugins/ptp_operator/ptp_operator_plugin.go
```diff
@@ -424,16 +424,37 @@ func processPtp4lConfigFileUpdates() {
...
-						Role:     role,
+
+					// Check if interface section exists in config
+					if interfaceSection, exists := allSections[*ptpInterface]; exists {
+						// Try to preserve existing role if interface exists and port ID matches
+						if p, e := ptp4lConfig.ByInterface(*ptpInterface); e == nil && p.PortID == index+1 {
+							role = p.Role
+						} else {
+							// Assign role based on masterOnly parameter from config
+							if masterOnlyValue, hasMasterOnly := interfaceSection["masterOnly"]; hasMasterOnly {
```
josephdrichard, line 435: What about if config is serverOnly?

aneeshkp: update to check for both clientOnly and serverOnly

plugins/ptp_operator/metrics/manager.go
aneeshkp: fixed

plugins/ptp_operator/ptp_operator_plugin.go
```diff
@@ -424,16 +424,90 @@ func processPtp4lConfigFileUpdates() {
...
+										} else {
+											// Default to SLAVE if no other role is specified
+											role = ptpTypes.SLAVE
+										}
+									} else {
+										// Default to SLAVE if no global section or clientOnly parameter
+										role = ptpTypes.SLAVE
+									}
+								}
+							} else if slaveOnlyValue, hasSlaveOnly := interfaceSection["slaveOnly"]; hasSlaveOnly {
```
josephdrichard, line 465: slaveOnly is not configurable on a per-interface basis

PR #597, opened 2025-09-25
basic support for chronyd

josephdrichard:
Add support for setting OS clock sync state based on information from chronyd.  More specifically, when chronyd is controlling the OS clock, then the OS clock sync state should be reported as locked.

### DIFF
--- a/plugins/ptp_operator/metrics/metrics.go
+++ b/plugins/ptp_operator/metrics/metrics.go
@@ -15,10 +15,12 @@ import (
 )
 
 var (
-	configFileRegEx = regexp.MustCompile(`(ptp4l|ts2phc|phc2sys|synce4l)\.[0-9]*\.config`)
+	configFileRegEx = regexp.MustCompile(`(ptp4l|ts2phc|phc2sys|synce4l|chronyd)\.[0-9]*\.config`)
 	// NodeName from the env
-	ptpNodeName        = ""
-	masterOffsetSource = ""
+	ptpNodeName          = ""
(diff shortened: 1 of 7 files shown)


plugins/ptp_operator/metrics/metrics.go
```diff
@@ -150,6 +153,14 @@ func (p *PTPEventManager) ExtractMetrics(msg string) {
 		p.ParseTBCLogs(processName, configName, output, fields, ptpStats)
 	case syncE4lProcessName:
 		p.ParseSyncELogs(processName, configName, output, fields, ptpStats)
+	case chronydProcessName:
+		ptpStats.CheckSource(ClockRealTime, configName, processName)
+		ptpStats[ClockRealTime].SetProcessName(processName)
+		if chronydValidSource.MatchString(output) {
+			p.GenPTPEvent(profileName, ptpStats[ClockRealTime], ClockRealTime, int64(0), ptp.LOCKED, ptp.OsClockSyncStateChange)
```
jzding, line 160: should ptp metrics be updated too?

josephdrichard: where would that be updated?

aneeshkp: similar to process is running or not. 
Also in linuxptpdaemon when you stopping phc2sys process,  I think it  will update process down in the log, but then you should do similar to chronyD (steop|start)

jzding (approved):
/lgtm

PR #657, opened 2026-03-20
[release-4.21] OCPBUGS-77871: Fix clock class metrics not restored after sidecar restart

jzding:
When cloud-event-proxy restarts, TriggerLogs() was called before the socket listener was started, so the linuxptp-daemon had nowhere to reconnect when re-emitting metrics. Remove the startup TriggerLogs() call and instead trigger logs in processMessages() when a new socket connection is accepted, which handles both initial startup and reconnection scenarios.


nocturnalastro (approved):
/lgtm

PR #659, opened 2026-03-23
OCPBUGS-78552: Generic sanity check for metrics

lack:
Adds CheckMetricSanity to prevent any metric update from being emitted
with an empty process or interface name. This ensures phantom metrics
are rejected and provides a stack trace warning to debug the caller.

Assisted-by: gemini-2.5-pro
Signed-off-by: Jim Ramsay <jramsay@redhat.com>


jzding (approved):
/lgtm

PR #660, opened 2026-03-24
update OWNERS

jzding:


examples/manifests/kustomization.yaml
```diff
@@ -10,5 +10,5 @@ replicas:
   name: cloud-consumer-deployment
 images:
 - name: cloud-event-consumer
-  newName: quay.io/redhat-cne/cloud-event-consumer
+  newName: quay.io/jacding/cloud-event-consumer
```
nocturnalastro: You need to revert this

jzding: yes thanks

OWNERS_ALIASES
```diff
@@ -2,18 +2,19 @@
...
 aliases:
   cloud-event-proxy-reviewers:
-    - aneeshkp
-    - atyronesmith
     - josephdrichard
     - jzding
+    - lack
     - nocturnalastro
     - rdavid
+    - SchSeba
```
nocturnalastro: Should this be sebsoto?

jzding: ah, good eye. fixed this for all repos.

rdavid (approved):

PR #663, opened 2026-04-10
Fix EventOutCh deadlock by making publish non-blocking

jzding:
PublishEventViaAPI sends to EventOutCh with a blocking channel send. When ProcessOutChannel (the consumer) is slow due to HTTP POST failures (e.g., unreachable subscribers), the 100-slot buffer fills up and the next publish blocks forever. This deadlocks the processMessages goroutine, freezing socket I/O and causing cloud-event-proxy to hang silently.

Fix by using select with default to drop events when the channel is full instead of blocking indefinitely. Dropped events are logged as warnings and counted as failures in metrics.

### DIFF
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -45,6 +45,20 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+const sendTimeout = 5 * time.Second
+
+// SendToChannel sends data to a channel with a timeout.
+// Returns true if the send succeeded, false if the channel was full for the
+// entire timeout duration.
+func SendToChannel(ch chan *channel.DataChan, d *channel.DataChan) bool {
+	select {
+	case ch <- d:
+		return true
+	case <-time.After(sendTimeout):
+		return false
+	}
+}
+
 // TransportType  defines transport type supported
 type TransportType int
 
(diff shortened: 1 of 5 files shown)


pkg/common/common.go
```diff
@@ -309,18 +309,21 @@ func PublishEvent(scConfig *SCConfiguration, e ceevent.Event) error {
...
 			Type:     channel.EVENT,
 			Status:   channel.NEW,
 			Data:     ceEvent,
 			Address:  resourceAddress, // this is the publishing address
 			ClientID: scConfig.ClientID(),
+		}:
+			log.Debugf("event source %s sent to queue to process", ceEvent.Source())
+			log.Debugf("event sent %s", cneEvent.JSONString())
+			localmetrics.UpdateEventPublishedCount(ceEvent.Source(), localmetrics.SUCCESS, 1)
+		default:
```
nocturnalastro: Can you set them to retry with a back-off instead of failing out right? 
Then perhaps fail it its over a certain time?

jzding: updated. added a 5 second timeout.

nocturnalastro (approved):
/lgtm

PR #667, opened 2026-04-22
OCPBUGS-81591: Fetch aliases before listening to socket

nocturnalastro:
This stops aliases defaulting to the port when cloud-event-proxy restarts but linuxptp-daemon is already running


jzding (approved):
/lgtm

PR #669, opened 2026-04-28
OCPBUGS-83551: Fix tbc startup ordering

nocturnalastro:
This PR fixes a startup race condition in the cloud-event-proxy PTP plugin where T-BC (Transparent Boundary Clock) configs were registered with ProfileType=NONE because loadInitialPtp4lConfigs() ran before the configmap was loaded and TBCProfiles populated. This caused spurious OsClockSyncStateChange FREERUN events on SLAVE-to-FAULTY transitions.

### DIFF
--- a/plugins/ptp_operator/filesystem/mock.go
+++ b/plugins/ptp_operator/filesystem/mock.go
@@ -1,0 +1,174 @@
+package filesystem
+
+import (
+	"fmt"
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+// MockFileSystem is a mock implementation of FileSystemInterface for testing.
+// Modeled after linuxptp-daemon's addons/intel/mock_test.go MockFileSystem.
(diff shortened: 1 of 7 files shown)


plugins/ptp_operator/filesystem/mock.go
```diff
@@ -0,0 +1,174 @@
+package filesystem
```
jzding, line 1: Without a _test.go suffix this would be shiped in production binaries. I suggest rename it to mock_test.go.

jzding:
@nocturnalastro could you add ticket number (story or bug) to the title and add a PR description?

plugins/ptp_operator/filesystem/mock.go
nocturnalastro: Can't unfortunately because the mock is used by the test of other packages.

jzding (approved):
/lgtm

PR #683, opened 2026-05-20
OCPBUGS-85092: handle CMD LIVE_START marker and add debug observability

edcdavid:
## Root Cause

See linuxptp-daemon PR: https://github.com/k8snetworkplumbingwg/linuxptp-daemon/pull/216

After a PTP reconfiguration, replay data (port roles, clock sync state) and live ptp4l data race on separate socket connections into CEP with no ordering guarantee. A stale replayed port role transition (e.g. `SLAVE→FAULTY`) processed after live data causes CEP to transition `openshift_ptp_clock_state` to HOLDOVER/FREERUN permanently — because port roles gate the clock state machine (LOCKED requires SLAVE), and port role events only fire on actual ptp4l state transitions, so there is no corrective event.

The daemon-side live gate (linuxptp-daemon PR) guarantees replay completes before live data is written. This CEP PR handles the new `CMD LIVE_START` marker that the daemon sends after the gate opens.

## Changes

- **`CMD LIVE_START` handling**: recognized in the scanner loop and skipped (`continue`) — not passed to `ExtractMetrics`. Provides a clear demarcation between replay and live data phases in logs.
- **`TriggerLogs` remains async**: the comment explains why synchronous would deadlock — the /emit-logs response writes replay data through the EventHandler's socket connection, which CEP accepts as a separate `processMessages` goroutine. If TriggerLogs blocks here, that goroutine's reader is stuck, the kernel buffer fills, and the emit handler blocks. The daemon's liveGate independently ensures no live data flows until replay completes.
- **Debug logging**: added throughout `processMessages`, `TriggerLogs`, `ExtractMetrics`, and `GenPTPEvent` at debug level for observability without changing default verbosity.
- **Mock server binds to loopback**: `startMockLogsServer` uses `127.0.0.1:8081` instead of `:8081` to avoid port contention and unintended external exposure in tests.

## Unit Tests Added

- `TestProcessMessages_LiveStartSkipped` — CMD LIVE_START is consumed without passing to ExtractMetrics
- `TestProcessMessages_LiveStartBetweenRegularMessages` — LIVE_START interspersed with normal ptp4l/phc2sys lines is handled correctly
- `TestProcessMessages_TriggerLogsFailureNonBlocking` — async TriggerLogs failure does not block message processing
- `TestProcessMessages_MultipleLiveStartOnly` — repeated LIVE_START messages are all safely skipped
- `TestLiveStartCommand_ConstantValue` — CMD LIVE_START is distinct from CMD RESTART and uses CMD prefix
- `TestProcessMessages_AliasReadyGating` — processMessages blocks until aliasReady is closed, then processes normally
- `Test_ExtractMetrics/ReplayThenLive_ClockStateCorruption` — proves stale replay port role permanently poisons clock state (documents why the gate is essential)
- `Test_ExtractMetrics/ReplayThenLive_PortRoleCorruption` — proves stale replay FAULTY overwrites live SLAVE role permanently
- `Test_ExtractMetrics/WithLiveGate_ReplayThenLive_MetricsCorrect` — with correct ordering (replay first, then live), metrics recover to LOCKED/SLAVE

## Depends On

- https://github.com/k8snetworkplumbingwg/linuxptp-daemon/pull/216

### DIFF
--- a/plugins/ptp_operator/metrics/manager.go
+++ b/plugins/ptp_operator/metrics/manager.go
@@ -431,6 +431,8 @@ func (p *PTPEventManager) GenPTPEvent(ptpProfileName string, oStats *stats.Stats
 	}
 
 	lastClockState := oStats.LastSyncState()
+	log.Debugf("GenPTPEvent: profile=%s resource=%s offset=%d clockState=%s lastState=%s eventType=%s",
+		ptpProfileName, eventResourceName, ptpOffset, clockState, lastClockState, eventType)
 	threshold := p.PtpThreshold(ptpProfileName, false)
 	switch clockState {
 	case ptp.LOCKED:
@@ -827,12 +829,15 @@ func (p *PTPEventManager) SetInitalMetrics() {
 }
 
 func (p *PTPEventManager) TriggerLogs() error {
+	log.Debugf("TriggerLogs: calling %s", logsEndpoint)
 	client := &http.Client{Timeout: 5 * time.Second}
 	resp, err := client.Get(logsEndpoint)
 	if err != nil {
+		log.Debugf("TriggerLogs: request failed: %v", err)
 		return err
 	}
 	defer resp.Body.Close()
+	log.Debugf("TriggerLogs: response status=%s", resp.Status)
 	return nil
 }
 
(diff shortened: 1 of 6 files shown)


jzding (approved):
/lgtm

PR #696, opened 2026-06-09
OCPBUGS-87883: Fix BC stuck on holdover

vitus133:
Currently holdover timer is not started if the corresponding interface can't be associated with a PTP profile. When a TR is lost, the holdover state is published as event and set in metrics, but the timer start is skipped due to the unresolved ptpOpts.
1. Fix nil PtpProcessOpts preventing holdover timer recovery. Populate process options after config file create and resolve profile lookups on holdover entry so the timer always starts even when the one-shot configmap watcher missed the node profile at startup.
2. Ensure the holdover goroutine always starts after HOLDOVER state is published and set in metrics. Close orphaned timer channels on reset. Add lifecycle logging for cluster diagnosis.

### DIFF
--- a/plugins/ptp_operator/config/config.go
+++ b/plugins/ptp_operator/config/config.go
@@ -141,7 +141,9 @@ func (pt *PtpClockThreshold) SafeClose() (justClosed bool) {
 	}()
 	select {
 	case <-pt.Close:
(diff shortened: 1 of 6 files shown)


nocturnalastro (approved):
/lgtm

PR #706, opened 2026-06-18
CNF-16778: Add unit test for ts2phc freerun clock state on high offset

yaronh12:
Adds a unit test verifying that when ts2phc reports s2 (LOCKED) but the
offset exceeds the configured threshold, the clock_state metric is
correctly overridden to FREERUN.
This covers the fix introduced in #321.


jzding (approved):
/lgtm

PR #707, opened 2026-06-21
OCPBUGS-87883: Cleanup clock_class metrics on ptpconfig deletion

vitus133:
Delete openshift_ptp_clock_class series when boundary-clock configs are removed via file watcher or zero-profile config map update.


nocturnalastro (approved):
/lgtm

PR #712, opened 2026-06-30
OCPBUGS-91644: Fix NTP-failover dual-publisher for CLOCK_REALTIME

vitus133:
In NTP-failover profiles (chronyd + phc2sys configured), phc2sys process-down was emitting a spurious OsClockSyncStateChange FREERUN event for CLOCK_REALTIME, conflicting with chronyd which is the sole E3 authority in NTP mode. This caused:
- Double CLOCK_REALTIME values in os-clock-sync-state events
- Stale openshift_ptp_clock_state{process="phc2sys"} metric

Fix: skip the OsClockSyncStateChange emission in processDownEvent when ChronydEnabled() is true for the profile. Also update the Prometheus gauge on the non-chronyd path to prevent stale metrics.


nocturnalastro (approved):
/lgtm

PR #701, opened 2026-06-11
OCPBUGS-88369: Fix CLOCK_REALTIME dual-publisher conflict in ParseTBCLogs

edcdavid:
### Commit 1: Fix CLOCK_REALTIME dual-publisher conflict in ParseTBCLogs

- **Removes the CLOCK_REALTIME override from `ParseTBCLogs`**: when the T-BC FSM transitions to FREERUN, `ParseTBCLogs` was unconditionally forcing `CLOCK_REALTIME` to FREERUN (with sentinel offset `-9999999999999999`), overriding `phc2sys` which may still be reporting LOCKED with a healthy offset. This dual-publisher conflict causes `CLOCK_REALTIME` to oscillate between LOCKED and FREERUN, or stay permanently stuck at FREERUN.
- **`CLOCK_REALTIME` is now managed exclusively by `phc2sys`**: when the T-BC chain loses its upstream source, the PHC will eventually drift, `phc2sys` will detect the growing offset, and `CLOCK_REALTIME` will transition to FREERUN naturally through the threshold-based mechanism in `extractRegularMetrics`.
- **Adds regression test** `TestTBCFreerunDoesNotOverrideClockRealtime` verifying that T-BC FREERUN does not emit `OsClockSyncStateChange` or alter `CLOCK_REALTIME` state.

### Commit 2: E1-aware E3 derivation per O-RAN O-Cloud API v04.00 Table 37

- **Adds E1-aware syncState downgrade in `ExtractMetrics`** (`metrics.go`): before calling `GenPTPEvent` for `CLOCK_REALTIME`, the phc2sys-derived syncState is now downgraded using `worst_of(phc2sys_state, E1_state)`. If E1 (PTP Lock State) is `FREERUN` or `HOLDOVER`, E3 cannot be `LOCKED` even when phc2sys offset is within threshold. T-GM profiles are skipped (they use the existing `lastOverallGMState` mechanism).
- **Removes `maybePublishOSClockSyncStateChangeEvent`** (`ptp4lParse.go`): this redundant third publisher of E3 events is now obsolete. The phc2sys path in `ExtractMetrics` is the single E3 publisher. Also removes the unused `ptpOpts` parameter from `handleHoldOverState`.
- **Fixes `GenPTPEvent` HOLDOVER-to-FREERUN gating** (`manager.go`): allows `HOLDOVER` -> `FREERUN` transitions for `OsClockSyncStateChange` events, since `GenPTPEvent` is now the sole E3 publisher.
[...]

PR #714, opened 2026-07-01
OCPBUGS-88369: Fix E1-aware E3 guard race in T-BC profiles

edcdavid:
## Summary

Follow-up to #701 (E1-aware E3 derivation). The `masterOffsetSource` guard that gated the E1 check (`masterOffsetSource != ts2phcProcessName`) introduced a race condition in T-BC profiles:

- In T-BC profiles, both `ptp4l` and `ts2phc` write to `ptpStats["master"].ProcessName`, causing `masterOffsetSource` to toggle between `"ptp4l"` and `"ts2phc"` on every log line.
- When `masterOffsetSource` happened to be `"ts2phc"`, the E1 check was skipped entirely, producing incorrect LOCKED E3 events while the upstream PHC was in HOLDOVER or FREERUN.

### Fix

Remove the guard. `GetMainClockName()` already returns the correct stats key per profile type (T-BC→"T-BC", T-GM→"GM", OC/BC→"master"), so the E1 check works correctly for all profiles without needing to inspect `masterOffsetSource`.

This also extends E1-aware E3 derivation to T-GM profiles: when the GM is in HOLDOVER, E3 is now correctly reported as HOLDOVER per O-RAN O-Cloud API v04.00 Table 37 row 2 (previously only FREERUN was handled via the `lastOverallGMState` mechanism).

### Changes

- `metrics.go`: Remove `masterOffsetSource != ts2phcProcessName` guard around E1 check, update comment
- `tbc_test.go`: Add regression test `TestE1CheckFiresRegardlessOfMasterOffsetSource` that reproduces the race by forcing `masterOffsetSource = "ts2phc"` and verifying E3 is still downgraded
- `metrics_test.go`: Update `lastOverallGMState` HOLDOVER test to expect HOLDOVER (correct per O-RAN Table 37 row 2) instead of LOCKED

## Test plan

- [x] All existing tests pass (`go test ./plugins/ptp_operator/metrics/...`)
- [x] New regression test verifies E1 check fires even when masterOffsetSource is "ts2phc"
- [x] `goimports` and `go vet` clean

Assisted-By: Cursor

### DIFF
--- a/plugins/ptp_operator/metrics/metrics.go
+++ b/plugins/ptp_operator/metrics/metrics.go
@@ -271,15 +271,13 @@ func (p *PTPEventManager) ExtractMetrics(msg string) {
 				// phc2sys is the single publisher — but if the upstream PHC is not
 				// traceable (E1 != LOCKED), E3 cannot be LOCKED even when
 				// phc2sys offset is within threshold.
-				// Skip for T-GM profiles (masterOffsetSource == ts2phc): T-GM uses
-				// the lastOverallGMState mechanism above instead.
-				if masterOffsetSource != ts2phcProcessName {
-					mainClockKey := ptpStats.GetMainClockName()
-					if e1Stat, ok := ptpStats[mainClockKey]; ok {
-						e1State := e1Stat.LastSyncState()
-						if e1State == ptp.FREERUN || e1State == ptp.HOLDOVER {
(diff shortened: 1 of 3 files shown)


nocturnalastro (approved):
/lgtm

PR #717, opened 2026-07-05
OCPBUGS-90708: fix CurrentState aggregation

vitus133:
Fix CurrentState API handler producing duplicate entries with conflicting states (FREERUN from stale phc2sys + LOCKED from chronyd). Now, when chronyd is active for a profile, non-chronyd entries are excluded from the response and vice-versa.


nocturnalastro (approved):
/lgtm

PR #718, opened 2026-07-06
[release-4.22] OCPBUGS-88703: Fix E1-aware E3 guard race in T-BC profiles

nocturnalastro:
The masterOffsetSource guard (`masterOffsetSource != ts2phcProcessName`) used to gate the E1 check was unstable: in T-BC profiles, both ptp4l and ts2phc write to ptpStats["master"].ProcessName, causing masterOffsetSource to toggle between "ptp4l" and "ts2phc" on every log line. When it happened to be "ts2phc", the E1 check was skipped entirely, producing incorrect LOCKED E3 events while the upstream PHC was in HOLDOVER or FREERUN.

Remove the guard. GetMainClockName() already returns the correct stats key per profile type (T-BC→"T-BC", T-GM→"GM", OC/BC→"master"), so the E1 check works correctly for all profiles without needing to inspect masterOffsetSource.

This also extends E1-aware E3 derivation to T-GM profiles: when the GM is in HOLDOVER, E3 is now correctly reported as HOLDOVER per O-RAN O-Cloud API v04.00 Table 37 row 2 (previously only FREERUN was handled via the lastOverallGMState mechanism).

### DIFF
--- a/plugins/ptp_operator/metrics/metrics.go
+++ b/plugins/ptp_operator/metrics/metrics.go
@@ -271,15 +271,13 @@ func (p *PTPEventManager) ExtractMetrics(msg string) {
 				// phc2sys is the single publisher — but if the upstream PHC is not
 				// traceable (E1 != LOCKED), E3 cannot be LOCKED even when
 				// phc2sys offset is within threshold.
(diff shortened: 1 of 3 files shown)


jzding (commented):
/lgtm

PR #720, opened 2026-07-07
Update golangci-lint to v2 and migrate lint config.

nocturnalastro:
Align CI and local lint baselines so new issues are reported while existing findings stay ignored, and remove deprecated deadcode nolint tags.

Generated-by: Cursor


jzding (approved):
/lgtm

PR #722, opened 2026-07-15
CNF-25885: Widen PTP clock thresholds when NTP failover is enabled

sebsoto:
When a profile has the ntpfailover plugin with gnssFailover: true, the user has accepted NTP-level accuracy as a fallback. The default +/-100ns offset threshold is too strict for that tolerance — it would declare FREERUN on minor drift that the user considers acceptable.

Widen the default threshold to +/-1000ns when NTP failover is enabled. Explicit PtpClockThreshold on the profile still takes precedence.


nocturnalastro (approved):
/lgtm

PR #723, opened 2026-07-28
OCPBUGS-98669: Default missing T-BC E1 to FREERUN for E3 after reconfig

nocturnalastro:
Prefer the T-BC stats key for T-BC profiles and treat unset E1 as FREERUN so a lone phc2sys LOCKED sample cannot make OS-clock E3 LOCKED too early after CEP restart or profile apply.

Generated-by: Cursor

### DIFF
--- a/plugins/ptp_operator/metrics/manager.go
+++ b/plugins/ptp_operator/metrics/manager.go
@@ -655,8 +655,8 @@ func (p *PTPEventManager) GetProfileType(name string) ptp4lconf.PtpProfileType {
 // name (e.g. "ptp4l.0.config"). Returns NONE if the config is unknown.
 func (p *PTPEventManager) GetProfileTypeByConfigName(configName types.ConfigName) ptp4lconf.PtpProfileType {
 	p.lock.RLock()
+	defer p.lock.RUnlock()
 	cfg, ok := p.Ptp4lConfigInterfaces[configName]
-	p.lock.RUnlock()
 	if !ok || cfg == nil {
 		return ptp4lconf.NONE
 	}
--- a/plugins/ptp_operator/metrics/manager.go
+++ b/plugins/ptp_operator/metrics/manager.go
@@ -651,6 +651,23 @@ func (p *PTPEventManager) GetProfileType(name string) ptp4lconf.PtpProfileType {
 	return ptp4lconf.NONE
 }
 
+// GetProfileTypeByConfigName returns the cached ProfileType for a ptp4l config
+// name (e.g. "ptp4l.0.config"). Returns NONE if the config is unknown.
+func (p *PTPEventManager) GetProfileTypeByConfigName(configName types.ConfigName) ptp4lconf.PtpProfileType {
+	p.lock.RLock()
+	cfg, ok := p.Ptp4lConfigInterfaces[configName]
+	p.lock.RUnlock()
+	if !ok || cfg == nil {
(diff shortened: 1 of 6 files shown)


plugins/ptp_operator/metrics/manager.go
```diff
@@ -651,6 +651,23 @@ func (p *PTPEventManager) GetProfileType(name string) ptp4lconf.PtpProfileType {
 	return ptp4lconf.NONE
 }
 
+// GetProfileTypeByConfigName returns the cached ProfileType for a ptp4l config
+// name (e.g. "ptp4l.0.config"). Returns NONE if the config is unknown.
+func (p *PTPEventManager) GetProfileTypeByConfigName(configName types.ConfigName) ptp4lconf.PtpProfileType {
+	p.lock.RLock()
+	cfg, ok := p.Ptp4lConfigInterfaces[configName]
+	p.lock.RUnlock()
```
vitus133: Why not deferred unlock?
the cfg is accessed below without the lock and can be mutated on the way (for example in RefreshProfileTypes)

plugins/ptp_operator/metrics/metrics.go
```diff
@@ -273,13 +273,19 @@ func (p *PTPEventManager) ExtractMetrics(msg string) {
...
-					e1State := e1Stat.LastSyncState()
-					if e1State == ptp.FREERUN || e1State == ptp.HOLDOVER {
-						syncState = OverallState(syncState, e1State)
-					}
+				if p.GetProfileTypeByConfigName(types.ConfigName(configName)) == ptp4lconf.TBC {
+					mainClockKey = types.IFace(stats.TBCMainClockName)
+				}
+				// Missing or unset E1 defaults to FREERUN so a lone OS-clock
+				// LOCKED sample cannot make E3 LOCKED early after reconfig.
+				e1State := ptp.FREERUN
```
vitus133: Would it be safer to leave the old behavior for "everything not T-BC"?
Something like
```
  if p.GetProfileTypeByConfigName(types.ConfigName(configName)) == ptp4lconf.TBC {
      mainClockKey = types.IFace(stats.TBCMainClockName)
      e1State := ptp.FREERUN
      if e1Stat, ok := ptpStats[mainClockKey]; ok && e1Stat.LastSyncState() != "" {
          e1State = e1Stat.LastSyncState()
      }
      syncState = OverallState(syncState, e1State)
  } else if e1Stat, ok := ptpStats[mainClockKey]; ok {
      e1State := e1Stat.LastSyncState()
      if e1State == ptp.FREERUN || e1State == ptp.HOLDOVER {
          syncState = OverallState(syncState, e1State)
      }
  }
``` 
The concern is that HA profiles will be affected. 
 Current code:  HA profile, phc2sys reports LOCKED:
  1. ptpStats["master"] doesn't exist, so E1 check is skipped
  2. syncState stays LOCKED (from phc2sys) 
  3. Event published at line 300 with LOCKED

  New code:
  1. GetProfileTypeByConfigName returns NONE, mainClockKey is "master"
  2. ptpStats["master"] doesn't exist -> e1State is set to FREERUN
  3. syncState = OverallState(LOCKED, FREERUN) returns FREERUN
  4. Event published at line 300 with FREERUN

vitus133 (commented):
Added a couple of concerns, WDYT?

vitus133 (approved):
/lgtm

PR #725, opened 2026-07-28
OCPBUGS-98227: Fix phc2sys / chrony stale metric

vitus133:
Fix phc2sys / chrony stale metric by cleaning up the old metric, so there is only exactly one CLOCK_REALTIME metric on the system at all times

Assisted by Cursor


nocturnalastro (approved):
/lgtm

PR #726, opened 2026-08-06
OCPBUGS-104524: Fix openshift_ptp_threshold metric not refreshed after PtpConfig change

jzding:
## Summary
- `openshift_ptp_threshold` Prometheus metric stays at defaults after a runtime PtpConfig `ptpClockThreshold` update when CEP is enabled
- The metric was only `.Set()` in `startPtpConfigSync()` (1s timeout window at startup). The lazy-reload path in `EnsureProcessOptions()` updated the internal `EventThreshold` map but never refreshed the gauge.
- Add an `OnThresholdUpdate` callback to `LinuxPTPConfigMapUpdate`, invoked at the end of `UpdatePTPThreshold()`. The plugin registers this callback to set the Prometheus metric, so both the initial sync and lazy-reload paths update the gauge.

Fixes: OCPBUGS-104524

## Test plan
- [ ] All existing unit tests pass (`go test ./plugins/ptp_operator/...`)
- [ ] Deploy on a cluster with CEP enabled, modify `ptpClockThreshold` in PtpConfig, verify `openshift_ptp_threshold` metric reflects new values within 30s

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/plugins/ptp_operator/ptp_operator_plugin.go
+++ b/plugins/ptp_operator/ptp_operator_plugin.go
@@ -31,6 +31,7 @@ import (
 	"github.com/redhat-cne/sdk-go/pkg/event"
 
 	"github.com/redhat-cne/cloud-event-proxy/plugins/ptp_operator/alias"
+	ptpConfig "github.com/redhat-cne/cloud-event-proxy/plugins/ptp_operator/config"
 	"github.com/redhat-cne/cloud-event-proxy/plugins/ptp_operator/ptp4lconf"
 
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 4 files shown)


plugins/ptp_operator/ptp_operator_plugin.go
```diff
@@ -200,14 +201,6 @@ func startPtpConfigSync(timeout time.Duration, nodeName string, configuration *c
 			eventManager.PtpConfigMapUpdates.UpdatePTPThreshold()
 			eventManager.PtpConfigMapUpdates.UpdatePTPSetting()
 			eventManager.RefreshProfileTypes()
-			for key, np := range eventManager.PtpConfigMapUpdates.EventThreshold {
```
nocturnalastro, line 203: I don't understand how this is different in terms of the config map not existing. If anything isn't this happening earlier?

nocturnalastro (approved):
/lgtm

PR #731, opened 2026-08-07
OCPBUGS-105425: Publish os-clock FREERUN when phc2sys stops selecting CLOCK_REALTIME

edcdavid:
## Summary
- Detect when phc2sys (`phc2sysOpts: -a -r`) stops selecting `CLOCK_REALTIME` as a sink after upstream loss (selection window: `already selected` / settle silence)
- Publish `OsClockSyncState` FREERUN so E3 does not stay stuck LOCKED after `CLOCK_REALTIME phc offset` samples stop
- Does **not** implement reverse sync (`-r -r` / `sys offset` / as-source); that mode is out of scope
- Does not bring back T-BC forcing E3 (removed in OCPBUGS-88369)

Fixes: OCPBUGS-105425

## Why OCPBUGS-88369 missed this
88369 stopped T-BC from forcing the OS clock FREERUN. After that, E3 only updated when phc2sys still printed `CLOCK_REALTIME phc offset` lines. With `-a -r`, after upstream loss those lines often stop (CLOCK_REALTIME is no longer selected as sink — not reverse sync, which needs `-r -r`), so E3 could stay LOCKED. Test T-BC often uses `-s <iface>` (samples keep coming), and the 88369 tests always fed another CLOCK_REALTIME sample — so this case was never hit.

Two rules that both must stay true:
1. T-BC FREERUN alone must not force E3 FREERUN.
2. When phc2sys stops disciplining CLOCK_REALTIME and samples stop, E3 must still go FREERUN.

Do not bring back “T-BC forces E3” to fix (2) — that breaks (1) again.
Covered together by the combined T-BC FREERUN + selection-silence unit test.

## Test plan
- [x] Unit tests for selection-window FREERUN / recovery / chronyd skip / T-BC+silence
- [ ] E2E BC upstream-loss: OsClockSyncState FREERUN then LOCKED on recovery (ptp-operator serial)

Assisted-By: Cursor

### DIFF
--- a/plugins/ptp_operator/metrics/os_clock_discipline.go
+++ b/plugins/ptp_operator/metrics/os_clock_discipline.go
@@ -1,0 +1,171 @@
+package metrics
+
+import (
+	"slices"
+	"strings"
+
+	"github.com/redhat-cne/cloud-event-proxy/plugins/ptp_operator/stats"
+	"github.com/redhat-cne/sdk-go/pkg/event/ptp"
+	log "github.com/sirupsen/logrus"
+)
+
+// OS-clock discipline (phc2sysOpts: -a -r)
+//
+// After upstream loss, phc2sys may stop selecting CLOCK_REALTIME as a sink, so
+// "CLOCK_REALTIME phc offset" samples stop and E3 can stay LOCKED. Watch the
+// selection burst after "reconfiguring after port state change":
+//   - NIC selected, CLOCK_REALTIME not (already selected / settle silence) → E3 FREERUN
+//   - selecting CLOCK_REALTIME or CLOCK_REALTIME phc offset → clear sticky, normal path
+//
+// Not reverse sync (-r -r). Do not force E3 from T-BC alone (OCPBUGS-88369).
+
+// osClockLineClass classifies a phc2sys log line for the selection-window
+// state machine.  Merging the per-line checks into one classifier keeps the
+// handler a clean switch.
+type osClockLineClass int
+
+const (
+	osClockLineOther       osClockLineClass = iota // no match — continue to normal processing
+	osClockLineReconfigure                         // "reconfiguring after port state change"
+	osClockLineWaiting                             // "source clock not ready" / "postponing" / "no PHC ready"
+	osClockLineSourceClock                         // "as domain source clock" / "as out-of-domain source clock"
+	osClockLineFinalised                           // "already selected" or " offset "
+)
+
+func classifyOsClockLine(output string) osClockLineClass {
+	if strings.Contains(output, "reconfiguring after port state change") {
+		return osClockLineReconfigure
+	}
+	if strings.Contains(output, "source clock not ready, waiting") ||
+		strings.Contains(output, "multiple source clocks available, postponing sync") ||
+		strings.Contains(output, "no PHC ready, waiting") {
+		return osClockLineWaiting
+	}
+	if strings.Contains(output, "as domain source clock") ||
+		strings.Contains(output, "as out-of-domain source clock") {
+		return osClockLineSourceClock
+	}
+	if strings.Contains(output, "already selected") ||
+		strings.Contains(output, " offset ") {
+		return osClockLineFinalised
+	}
+	return osClockLineOther
+}
+
+func (p *PTPEventManager) handlePhc2sysSyncDirection(profileName, configName, output string,
+	fields []string, ptpStats stats.PTPStats) bool {
+	ptpStats.CheckSource(ClockRealTime, configName, phc2sysProcessName)
+	cr := ptpStats[ClockRealTime]
+
+	onSettle := func(generation uint64) {
+		p.extractMu.Lock()
+		defer p.extractMu.Unlock()
+		if cr.OsClock.FinaliseIfGeneration(generation) {
+			log.Infof("os-clock: profile=%s reason=selection-settle → E3 FREERUN", profileName)
+			p.publishOsClockFreerunUndisciplined(profileName, ptpStats)
+		}
+	}
+
+	switch classifyOsClockLine(output) {
+	case osClockLineReconfigure:
+		cr.OsClock.Begin()
+		log.Debugf("os-clock: profile=%s selection-window=begin", profileName)
+		return true
+
+	case osClockLineWaiting:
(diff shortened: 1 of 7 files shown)


plugins/ptp_operator/stats/stats.go
```diff
@@ -45,6 +46,19 @@ type Stats struct {
...
+	// system clock from the network PHC. While true, os-clock-sync-state
+	// should be FREERUN because CLOCK_REALTIME is not network-disciplined.
+	osClockReverseSync bool
+	// Selection window state machine (see docs/REVERSE_SYNC_DETECTION.md):
+	//   begin → record selections → terminal | settle | abort
+	// activateReverse = sawOther && !sawClockRealtime.
+	phc2sysSelectionMu               sync.Mutex
+	phc2sysSelectionPending          bool
+	phc2sysSawClockRealtimeSelection bool
+	phc2sysSawOtherSelection         bool
```
nocturnalastro: These together seem like a state. Do they really need to all be bools or can we just have an enum.

nocturnalastro: Also feels like this  whole thing is its own small object that perhaps could sit on stats but maybe could just be on the event manager. Its like its own little bit of state.

edcdavid: Fixed. I left the changes in the stats package because it is already per profile/less refactor.

plugins/ptp_operator/metrics/os_clock_discipline.go
```diff
@@ -0,0 +1,268 @@
...
+// CLOCK_REALTIME as a sink and stops emitting "CLOCK_REALTIME phc offset" samples.
+// Without detection, os-clock-sync-state (E3) can stay LOCKED indefinitely.
+// This is not reverse sync (that requires -r -r); reverse sync is out of scope.
+//
+// Approach: accumulate a short selection window after
+// "reconfiguring after port state change", then decide:
+//   - CLOCK_REALTIME (or "system clock") selected → still disciplined (clear sticky)
+//   - only NIC selected + terminal/settle → undisciplined → E3 FREERUN
+//
+// "Terminal" means a phc2sys log line that ends the selection window (enough
```
nocturnalastro, line 24: Would it be better to use something like finalised? to avoid the need for a comment?

plugins/ptp_operator/stats/stats.go
```diff
@@ -197,6 +201,65 @@ func (s *Stats) SetLastSyncState(val ptp.SyncState) {
...
+func (s *Stats) SawPhc2sysOtherSelection() bool {
+	return s.phc2sysSync.SawOtherSelection()
+}
+
+// ArmPhc2sysSelectionSettle arms a settle timer used when phc2sys ends a reverse
+// selection with silence (single-NIC: only "selecting <iface> for synchronization").
+// onSettle runs on a timer goroutine with the window generation at arm time; it
+// must call FinalizePhc2sysSelectionWindowIfGeneration and publish FREERUN when
+// appropriate.
+func (s *Stats) ArmPhc2sysSelectionSettle(delay time.Duration, onSettle func(generation uint64)) {
```
nocturnalastro, line 242: Is this a test helper if so it should probably go in a `_test.go` file but with `package stats` so that its only adding when running tests

plugins/ptp_operator/metrics/os_clock_discipline.go
```diff
@@ -0,0 +1,268 @@
...
+func (p *PTPEventManager) beginPhc2sysSelectionWindow(profileName string, cr *stats.Stats) bool {
+	cr.BeginPhc2sysSelectionWindow() // clears prior selections + cancels any settle timer
+	log.Debugf("os-clock: profile=%s selection-window=begin", profileName)
+	return true
+}
+
+// abortPhc2sysSelectionWindow drops an open window without publishing FREERUN
+// (waiting/postponing: phc2sys has not finished choosing). Still returns true
+// so the waiting line is consumed.
+func (p *PTPEventManager) abortPhc2sysSelectionWindow(profileName string, cr *stats.Stats, reason string) bool {
```
nocturnalastro, line 98: This doesn't use `p` does it make sense for it to be a method on PTPEventManager? 
There are a few like this.

plugins/ptp_operator/metrics/os_clock_discipline.go
```diff
@@ -0,0 +1,268 @@
...
+		// running; a prior settle must not Finalize a window opened by a later Begin.
+		if cr.FinalizePhc2sysSelectionWindowIfGeneration(generation) {
+			log.Infof("os-clock: profile=%s reason=selection-settle → E3 FREERUN", profileName)
+			p.publishOsClockFreerunUndisciplined(profileName, ptpStats)
+		}
+	})
+}
+
+// SetPhc2sysSelectionSettleDelayForTest overrides the settle delay (tests only).
+func SetPhc2sysSelectionSettleDelayForTest(d time.Duration) (restore func()) {
```
nocturnalastro, line 205: Test help should be in `_test.go` file as an internal test.

plugins/ptp_operator/metrics/os_clock_discipline.go
```diff
@@ -0,0 +1,268 @@
...
+}
+
+// isPhc2sysSelectionTerminal reports whether output is a selection-window
+// terminal: a log line that closes the open window so we can decide whether
+// CLOCK_REALTIME remains a sink. Today that is "already selected" (dual-port
+// same-PHC end of burst) or any " offset " sample. CLOCK_REALTIME offsets are
+// still matched here so finalizePhc2sysSelectionTerminal can close the window
+// as disciplined. Mid-burst "selecting <iface>…" lines are not terminals;
+// silence without a terminal is handled by the settle timer.
+func isPhc2sysSelectionTerminal(output string) bool {
```
nocturnalastro, line 245: These seem like they should just one function that returns the Phase.

nocturnalastro: We're spreading logic in lost of different places ATM.

Perhaps even the phys2sys Sync Direction object should just take the log line in and handle the state transition? Can always pass PTPEventManager into the update function so it can handle calling the publish event.

nocturnalastro (approved):
/lgtm

